PR #19, opened 2023-12-05
feat: add toggling status effects from ActorSheet

xeqi:
Add click handlers for the status effect icons that will toggle status effects in a similar way to the token hud.

Show status effect icons with medium transparency when not active, low transparency when hovered, and no transparency when active.

Refactor grid layout to use the status effect information from Config.

Temp


xeqi:
Ended up using v11 specific code.  Going to close this while I research how to have compatibility for both.

PR #41, opened 2024-01-14
Add bond and trait invocation

Shourn:
- refactor rolls
- add generic check dialog
- add reroll dialog
- add push dialog


templates/chat/chat-check.hbs
```diff
@@ -0,0 +1,22 @@
+{{log this}}
```
Wigginns, line 1: Leftover logging

Wigginns (commented):
Looking good. Feels like this'll pave the way for some other ideas. :) Haven't had a chance to try it yet. Noticed a leftover logging statement

PR #43, opened 2024-01-15
Weapon damage and accuracy

Wigginns:
Damage is calculated with multiple permutations of `Has Accuracy? Use Weapon Accuracy` and `Has Damage? Use Weapon Damage`

It follows these rules:

Use weapon damage if "Use Weapon Damage"
Use spell/ability damage if Not "Use Weapon Damage"
Add bonus spell/ability damage if "Use Weapon Damage" and useItem
Change type of damage if it's defined in the rollInfo.damage

Example of an infusion that uses the weapon accuracy, adds damage and changes the type:
image
image

Example of a Spellblade skill that uses weapon accuracy but does not use weapon damage:
image
image

Example of an infusion that does not use weapon accuracy, adds damage and changes the type:
image
image

Example regular weapon attack:
image
image

Changes in `actor-section-features.hbs`:

Before:
image

After:
image

### DIFF
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -251,8 +251,8 @@ export class FUItem extends Item {
 		const accCheckBonus = this.actor.system.bonuses.accuracy.accuracyCheck;
 		const accBonus = isWeapon ? accCheckBonus : magicCheckBonus || 0;
 
-		const attrs = isWeapon ? item.system.attributes : item.system.rollInfo.attributes;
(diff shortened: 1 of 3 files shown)


Wigginns:
I just noticed that the example `Example of an infusion that does not use weapon accuracy, adds damage and changes the type:` has the wrong accuracy on the chat card, working on a fix

Wigginns (commented):
Addressed in fc34ae8

Wigginns:
Large conflicts expected with https://github.com/League-of-Fabulous-Developers/FoundryVTT-Fabula-Ultima/pull/41 

Will rework from there.

PR #62, opened 2024-02-15
Feat : Translation improvement

Zakarik:
Hello, I'm a French-speaking developer.

I'm about to start using this system and I noticed that most of the selects in the item sheets weren't translated, so I took a look at it.

To simplify things, I used a handlebars directly integrated into Foundry to automatically generate the list of select options I could do in a simple way.
https://foundryvtt.com/api/classes/client.HandlebarsHelpers.html

To make everything fully translatable, the missing CONFIGs would have to be created.

I also added lines to two translation files - I don't know German and Italian to do it for these languages.

What I've added translates the items/actors in these lists:
image

You'll notice that I've integrated them twice, in two different ways. That's because in V11, the translation system for these lists has changed, so it's just a matter of maintaining V10 & V11 compatibility. It has no particular impact to have both in the translation file.


Shourn:
Looks good from my side. Thank you for your contribution!

Issue #146, opened 2024-08-15
'No Damage' spells showing damage.


4m2c4bl3:
With the 'Has Damage?' checkbox unchecked, damage still shows in roll. 
![image](https://github.com/user-attachments/assets/f88e2a46-ec39-4bec-86c6-cf1273465868)

it looks like the 'hasDamage' value/concept got lost in the transition to checks v2. I don't see it in the check-configuration.mjs CheckConfigurer class anywhere, or the checks-v2.mjs file at all.

4m2c4bl3:
@Shourn this seems to be back in the latest beta. I noticed my code got removed during the refactor you were doing

![Image](https://github.com/user-attachments/assets/9a61a8f4-8470-458d-b955-5f1dc9abbf2e)

Azurelol:
Should be fixed now.

PR #147, opened 2024-08-15
fix for not using hasDamage value in checks v2

4m2c4bl3:
#146 - Solving this bug I'm experiencing. Since I was opening so many I thought it only fair I try to help with them. I don't think this is ideal, it's not elegant, but there seems to be a lot of places where this logic can be injected, and a very complicated code base. If this isn't the best place to do it, then it's at least an idea of how to start? :)


spyrella:
Looks good to tart, if a problem arises, we could work from there.

PR #152, opened 2024-08-19
Fix: Twin Shield Tweaks

MattByName:
Hi there,

I made some tweaks to the Guardian Twin Shields ability as the book says that 'you gain the benefits of both items' (197) but in the compendium the bonus was only being applied once. I also added a Twin Shields item for when the player has one runic and one bronze shield

## Changes
- Twin Shields > set def bonus to 4
- Twin Shields (Runic) > set def to 4, set magic def to 4
- Twin Shields (Runic & Bronze) > set def to 4 and magic def to 2

Sorry if I've misunderstood the conventions on how you're building items. Thanks for all your hard work, the system is looking great.

### DIFF
--- a/src/packs/skills/weapon_Twin_Shields_2LyT2btVVPDPYY4h.json
+++ b/src/packs/skills/weapon_Twin_Shields_2LyT2btVVPDPYY4h.json
@@ -96,7 +96,7 @@
         {
           "key": "system.derived.def.bonus",
           "mode": 2,
-          "value": "2",
+          "value": "4",
           "priority": null
         }
       ],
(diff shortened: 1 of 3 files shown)


MattByName:
I just noticed I didn't update the uuid inside the new item json, and only put it in the filename. I'm not sure what the best way to generate an id is, I just made one in foundry by creating a copy of one of the other items. Will that cause problems?

spyrella:
No worries, i just generated a new id and edited the id in the json.

Christopher-R-Perkins:
I thought the idea was that you would have the twin shields item only replace one of the shields ands still have the other equipped on your offhand.

fultimator-official:
> I thought the idea was that you would have the twin shields item only replace one of the shields ands still have the other equipped on your offhand.

I guess the issue with this is that twin shield counts as a two handed weapon so it’ll take up both main and off hand. I’m holding off doing anything with shields until the equipment system is overhauled, where equipment data is stored in the actor instead of items. From there we could work on equipment validation.

Maybe adding a checkbox that’ll allow you to equip two handed weapons in one slot. So you could equip another shield and gain their benefits.

spyrella:
To update everyone, in V2.4.0, with the new implementation of EquipmentHandlerV2, the other twin shields wouldn't be necessary anymore since you can now equip two shields and also equip the twin shield item into a "phantom slot".
image

PR #154, opened 2024-08-23
single target damage

4m2c4bl3:
👋 
I think this should solve the "Apply damage to individual targets" TODO item in your discord.


spyrella:
Much thanks!

PR #155, opened 2024-08-25
added skill list to combat section

aleszreis:
Attempt to fix issue where NPC skills and class skills/features couldn't be dragged directly from compendium to NPC sheet, since it only considered items classified as "Misc Ability". 

Added a "Skills" list to the "Combat" tab (actor-section-combat.hbs), reutilizing code from the actor-section-classes.hbs skill section.


spyrella:
I'll note that "Skills" item type specially refers to pc skills. I'll just leave it there until I create a dedicated npcSkills item type.

PR #156, opened 2024-08-25
feat: damage hooks

KSops:
Adds two hooks to the damage pipeline.
'projectfu.damage.beforeApply' => altering overall values just before any meaningful calculations start
'projectfu.damage.applyTarget' => altering data per-target just before application
The hooks can modify the provided data structure and the changes will be picked up and propagated to the logic.

Refactors out duplicate code

Repairs and adds some JSDoc typedefs

Centralizes some hook constants for ease of lookup, enumeration, and hopefully documentation

### DIFF
--- a/module/helpers/apply-damage.mjs
+++ b/module/helpers/apply-damage.mjs
@@ -1,187 +1,179 @@
 import { FU, SYSTEM } from './config.mjs';
+import { FUHooks } from '../hooks.mjs';
 import { Flags } from './flags.mjs';
 import { ChecksV2 } from '../checks/checks-v2.mjs';
 import { CheckConfiguration } from '../checks/check-configuration.mjs';
 import { DamageCustomizer } from './damage-customizer.mjs';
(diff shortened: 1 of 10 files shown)


module/helpers/apply-damage.mjs
```diff
@@ -1,187 +1,179 @@
...
-						() => {
-							disabled = false;
-						},
-					);
-				} else {
-					handleDamageApplication(event, targets, {});
-				}
-			}
-		};
+	if (ChecksV2.isCheck(message)) {
```
KSops, line 58: Before, there were two executions. One for checks v1 and one for checks v2. I'm making the assumption here that the intent is that checks v1 and v2 are mutually exclusive. This could use a check on the behavioral change.

spyrella:
Looking pretty good, i'll be trying it out later tonight.

KSops:
I forgot the exports are referenced elsewhere like inline damage, and since the method signature changed it's broken. I'll need to patch that up

PR #180, opened 2024-12-06
Add support for improvised effect calculation on inline damage and resource gain/loss

Azurelol:
Adds support for using improvised effect into the following inline command ,where improvised effect:

$effect: `Minor|Heavy|Massive`

- `@DMG[$number $type]`,  `@DMG[$effect $type]`
- `@GAIN[$number $type]`  `@GAIN[$effect $type]`
- `@LOSS[$number $type]`,  `@LOSS[$effect $type]`

This allows having spells/skills scale to the level of the source within the following ranges:
- [5,19]
- [20,39]
- [40,INF]

The amount calculated from an improvised effect is a simple mapping between the level range and the effect "potency" where the level is calculated based on the owner of the item. This works equally for spells, skills and active effects. (When an active effect is dropped on an actor, that becomes its source)

Example:
image

### DIFF
--- a/module/helpers/improvised-effect.mjs
+++ b/module/helpers/improvised-effect.mjs
@@ -72,7 +72,9 @@ function calculateAmountFromContext(dataset, source, targets) {
 		level = source.system.level.value;
 	} else {
 		if (targets.length > 0) {
-			level = targets.reduce((a, b) => b.system.level.value > a.system.level.value);
+			level = targets.reduce((max, target) => {
+				return Math.max(max, target.system.level.value);
+			}, -Infinity);
 		} else {
 			console.warn(`No actor was given to determine level, thus used the default (5).`);
 		}


module/helpers/improvised-effect.mjs
```diff
@@ -0,0 +1,108 @@
...
+const effectIndex = {
+	minor: 0,
+	heavy: 1,
+	massive: 2,
+};
+
+/**
+ * Tier: Effect [Minor,Heavy,Massive]
+ */
+const amountPerTier = [
```
KSops, line 28: Isn't directly clear how this data structure is organized at first glance. Comment could be interpreted as the first index being minor/heavy/massive.

module/helpers/improvised-effect.mjs
```diff
@@ -0,0 +1,108 @@
...
+	} else {
+		if (targets.length > 0) {
+			level = targets.reduce((a, b) => b.system.level.value > a.system.level.value);
+		} else {
+			console.warn(`No actor was given to determine level, thus used the default (5).`);
+		}
+	}
+
+	const amount = ImprovisedEffect.calculateAmount(level, effect);
+	console.info(`Calculated amount for level ${level}, effect ${effect} = ${amount}`);
```
KSops, line 75: nit: console.debug might help reduce console clutter

module/helpers/improvised-effect.mjs
```diff
@@ -0,0 +1,108 @@
...
+	if (effect === undefined) {
+		return Number(dataset.amount);
+	}
+
+	let level = 5;
+	if (source !== undefined) {
+		level = source.system.level.value;
+	} else {
+		if (targets.length > 0) {
+			level = targets.reduce((a, b) => b.system.level.value > a.system.level.value);
```
KSops, line 68: This is intended to return the highest level of targets? Couldnt get the reduce to work in a minimal example

Azurelol: Yes, to take the highest level.

module/helpers/improvised-effect.mjs
Azurelol: Whoops. Sorry, did not mean to commit this here. I will remove.

Azurelol:
Sorry for the late follow-up; will address your feedback now.

module/helpers/improvised-effect.mjs
```diff
@@ -0,0 +1,108 @@
...
+ * Calculates the improvised amount for a given effect
+ * @param {*} dataset
+ * @param {FUActor} source
+ * @param {FUActor[]} targets
+ * @returns The amount as an integer, null otherwise
+ */
+function calculateAmountFromContext(dataset, source, targets) {
+	const effect = dataset.effect;
+	if (effect === undefined) {
+		return Number(dataset.amount);
```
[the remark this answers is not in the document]
Azurelol: Ack. Thanks, will fix.

module/helpers/improvised-effect.mjs
Azurelol: Good point. I will rewrite this comment for improved clarity.

module/helpers/improvised-effect.mjs
```diff
@@ -0,0 +1,108 @@
...
+ */
+
+/**
+ * @param {Number} level The character level
+ * @param {ImprovisedEffectType} effect The improvised effect type
+ * @returns
+ */
+function calculateAmount(level, effect) {
+	const tier = getCharacterTier(level);
+	const index = effectIndex[effect];
```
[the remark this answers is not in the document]
Azurelol: In this case I am okay with it throwing an exception as it should cause the text enricher to fail and leave the incorrectly formatted command to show as such for the user.

Azurelol:
Thank for your feedback, @KSops !

KSops:
Thanks for the great addition!

module/helpers/improvised-effect.mjs
KSops: I think something closer to this might achieve the max level (stolen from AI):
```
let maxValue = targets.reduce((max, target) => {
    return Math.max(max, target.system.level.value);
}, -Infinity); // Start with negative infinity to ensure any number is higher
```

Azurelol: Done!

PR #183, opened 2024-12-12
Add support for expressions in the inline chat commands

Azurelol:
Adds support  for evaluating expressions provided in the `Expressions` module, property referencing and function invocation in the inline chat commands such as `@DMG, @HEAL, @GAIN, @LOSS`

It adds support for the following:

- `@item|actor.<property>`
- `$<variable>` : These would be shorthands for very common usages like item level, which would be $il, can also support alias $sl
- `&<macro>`: for custom functions that the system can provide. The most beneficial one to provide would be the one that picks a value based on the character's level. I went with  &step(40,50,60) .
- `minor|heavy|massive` effect: Using the level of the actor to calculate the amount

### DIFF
--- a/module/documents/actors/character/character-data-model.mjs
+++ b/module/documents/actors/character/character-data-model.mjs
@@ -10,7 +10,6 @@ import { DerivedValuesDataModel } from '../common/derived-values-data-model.mjs'
 import { EquipDataModel } from '../common/equip-data-model.mjs';
 import { PilotVehicleDataModel } from './pilot-vehicle-data-model.mjs';
 import { SETTINGS } from '../../../settings.js';
-import { ImprovisedEffect } from '../../../helpers/improvised-effect.mjs';
 
 const CLASS_HP_BENEFITS = 5;
 const CLASS_MP_BENEFITS = 5;
@@ -137,25 +136,6 @@ export class CharacterDataModel extends foundry.abstract.TypeDataModel {
 	 */
 	get actor() {
 		return this.parent;
-	}
-
-	/**
-	 * @description Given 3 amounts, picks the one for this characters' level
-	 * @param {Number} first
-	 * @param {Number} second
-	 * @param {Number} third
-	 */
-	byLevel(first, second, third) {
-		const tier = ImprovisedEffect.getCharacterTier(this.level.value);
-		switch (tier) {
-			case 0:
-				return first;
-			case 1:
-				return second;
-			case 2:
-				return third;
-		}
-		return null;
 	}
 
 	prepareBaseData() {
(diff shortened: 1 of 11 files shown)


KSops:
Implementing this specifically for inline effects is fine for now, but I think eventually being able to handle dynamic expressions in other places will be beneficial and having a standardized DSL for the entire project would help.
So a utility module/class for dynamic evaluation that takes a string to parse and evaluate against a given context. And that way consumers can simply call an exposed interface call like `evaluate(value: string, context: object)` 
That way we can refence this common module for how the DSL works instead of worrying about specific contexts in which evaluation happens.

Azurelol:
> Implementing this specifically for inline effects is fine for now, but I think eventually being able to handle dynamic expressions in other places will be beneficial and having a standardized DSL for the entire project would help.
> So a utility module/class for dynamic evaluation that takes a string to parse and evaluate against a given context. And that way consumers can simply call an exposed interface call like `evaluate(value: string, context: object)` 
> That way we can refence this common module for how the DSL works instead of worrying about specific contexts in which evaluation happens.

Oh yeah, I definitely agree with providing a standard API for handling expression evaluation for the DSL in the system. I will go ahead and factor that out into its own class.

Azurelol:
It's been factored out and a bit simpler to reason about now that it's on its own module.

Azurelol:
One last question for the time being is to decide whether to use a different symbol for executing functions or use the same one as property referencing `@`. If we use different ones then we can invoke functions from referenced properties without having to deal with scoping issues. 

It may also be good to think of a symbol for short-hand syntactic sugar for things such as `item level`, `character level`, `target count`.

Azurelol:
After the latest round of changes I believe it's in a good state. I will await your review.

module/expressions/expressions.mjs
```diff
@@ -0,0 +1,272 @@
...
+	return expression.replace(pattern, evaluate);
+}
+
+/**
+ * @description Evaluates improvised effects
+ * @param {String} expression
+ * @param {ExpressionContext} context
+ * @returns {String}
+ */
+function evaluateEffects(expression, context) {
```
KSops, line 114: If you have special vars defined, I would move these seemingly special keywords there to cut down on special parsing/substitution cases

module/documents/actors/character/character-data-model.mjs
```diff
@@ -138,6 +139,25 @@ export class CharacterDataModel extends foundry.abstract.TypeDataModel {
 		return this.parent;
 	}
 
+	/**
+	 * @description Given 3 amounts, picks the one for this characters' level
+	 * @param {Number} first
+	 * @param {Number} second
+	 * @param {Number} third
+	 */
+	byLevel(first, second, third) {
```
KSops, line 148: nit: feels like convenience methods like this should be in a decorator of some sort or a static method that accepts an actor/character. keeping the core model lean might improve maintenance but I'm not an involved system dev as others

Azurelol: Agreed. I will move this method out of here onto the expressions module for the time since it will only be used by it.

module/expressions/expressions.mjs
Azurelol: I am on the fence doing so, particularly since the special variables will all most likely be abbreviated.

Azurelol: Though I could add support for the additional syntax.  Just need to sleep on whether to remove the old syntax now as well.

KSops (approved):
I didnt look too deeply on testing it so if it works it works :)

PR #187, opened 2024-12-23
Fix inline source usage for events

Azurelol:
Fixes the issue where the actor's methods could not be resolved for drag events.

It came to be due to non-primitives being passed in the `.dragstart` event, which get serialized then deserialized without type information. 

I had to refactor the events and the API involved so that we only pass in the uuids for the item/actor instead then resolve it at the very end.

### DIFF
--- a/module/expressions/expressions.mjs
+++ b/module/expressions/expressions.mjs
@@ -13,6 +13,25 @@ export class ExpressionContext {
 		this.item = item;
 		this.targets = targets;
 	}
+
+	/**
+	 * @param {String} actorUuid
+	 * @param {String} itemUuid
+	 * @param {FUActor[]} targets
+	 * @returns {ExpressionContext}
+	 */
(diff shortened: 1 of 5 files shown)


module/expressions/expressions.mjs
```diff
@@ -140,12 +167,12 @@ function evaluateVariables(expression, context) {
...
 			// Character level
 			case 'cl':
-				assertActorInContext(context, match);
+				context.assertActor(match);
 				return context.actor.system.level.value;
 			// Item level / skill level
 			case 'il':
 			case 'sl':
-				assertItemInContext(context, match);
+				context.assertActor(match);
```
KSops, line 175: assertItem* ?

Azurelol: Ack! Thank you for catching it.

KSops (approved):

spyrella (approved):

PR #189, opened 2024-12-25
Refactor the damage and resource pipelines

Azurelol:
Refactors the damage and resource pipelines to be standardized and to provide a sequence of steps

- Adds a new `pipelines` module, which now has the damage and resource pipelines
- The damage pipeline has been broken into sequence of discrete steps
- Adds support for using flags set by active effects for edge cases like `Zero Shield`
- Provides 3 hooks for the damage pipeline: before affinities, after affinities and at the very end
- Expands on the `BonusesDataModel`, `DamageBonusesDataModel` models to support skills such as `Defensive Mastery`, `Symbol of Weakness`.


Azurelol:
After this latest iteration I feel more confident on the architecture, so I believe it's worth taking a look and now and giving feedback in order to improve the design.

module/pipelines/damage-pipeline.mjs
```diff
@@ -0,0 +1,337 @@
...
+ */
+
+/**
+ * @type {Record<number, DamageModifier>}
+ */
+const affinityDamageModifier = {
+	[FU.affValue.vulnerability]: (damage) => damage * 2,
+	[FU.affValue.none]: (damage) => damage,
+	[FU.affValue.resistance]: (damage, { shift }) => (shift ? damage : Math.floor(damage / 2)),
+	[FU.affValue.immunity]: (damage, { shift, ctrl }) => (shift && ctrl ? damage : 0),
```
KSops, line 72: I was hoping we'd address the key modifiers on the next refactor.
I don't feel like modifier keys have any business being in logic layers. It's a UI-related issue and should be handled by the event consumer/hook. If in the future we somehow have other means of expressing "ignore resistance" that comes from a keyboard command rather than shift+click, we'd have to add more handler code in this logic layer.
The UI should be handled and the intent of "ignore resistance" is what should be passed to a pipeline imho.

Azurelol: I completely agree with you and would like to strip this out on the next pass. I think with a "traits" system we should be able to express intents in an uniform way that both UI and other procedures can use.

module/pipelines/damage-pipeline.mjs
```diff
@@ -0,0 +1,337 @@
...
+		);
+	}
+	return Promise.all(updates);
+}
+
+/**
+ * @param {?} message
+ * @param {jQuery} jQuery
+ */
+function onRenderChatMessage(message, jQuery) {
```
KSops, line 242: nit: Feels like it ought to be in some UI helper method instead of pipeline but I also don't have any concrete suggestions. Not a big deal since it's purely organizational

module/hooks.mjs
```diff
@@ -1,10 +1,43 @@
...
+	 * @description Add/modify the bonuses and modifiers before the result is calculated.
+	 * @example callback(context) { context.modifiers.set("foobar", 42); return true; }
+	 * @example callback(context) { context.modifiers.delete("affinity"); return true; }
+	 */
+	DAMAGE_PIPELINE_COLLECT: 'projectfu.pipelines.damage.collect',
+	/**
+	 * @description Modify the result after it's been calculated.
+	 * @example callback(context) { context.result += 42; return true; }
+	 */
+	DAMAGE_PIPELINE_CALCULATE: 'projectfu.pipelines.damage.calculate',
```
KSops, line 17: nit: naming might imply the step is for calculating. either change to past tense (calculated) or both steps to pre_calculate and post_calculate

module/pipelines/damage-customizer.mjs
```diff
@@ -1,264 +1,264 @@
...
-				}
-			},
-		},
-		{
-			width: 440,
-			classes: ['projectfu', 'unique-dialog', 'backgroundstyle'],
-		},
-	).render(true);
-}
+// import { FUActor } from '../documents/actors/actor.mjs';
```
KSops, line 1: leftover from refactor?

Azurelol: Ack, yes. Thanks.

module/hooks.mjs
Azurelol: Very well; I was on the fence too on which tense to use for these hooks. But we can go with past-tense convention for the system.

module/pipelines/damage-pipeline.mjs
```diff
@@ -0,0 +1,399 @@
...
+ * @param {DamageRequest} request
+ * @return {Promise<Awaited<unknown>[]>}
+ */
+async function process(request) {
+	if (!request.validate()) {
+		return Promise.reject('Request was not valid');
+	}
+
+	// TODO: Remove once users have migrated from legacy hooks
+	const beforeApplyHookData = new BeforeApplyHookData(request);
```
KSops, line 246: We don't use the result of the modified damage? The point of the hook is to be able to make changes to the calculation

module/hooks.mjs
```diff
@@ -1,10 +1,43 @@
...
+ * @remarks For most callbacks returning false will prevent anything further down the call-chain from being invoked.
+ */
 export const FUHooks = {
-	// Damage
+	/**
+	 * @description Add/modify the bonuses and modifiers before the result is calculated.
+	 * @example callback(context) { context.modifiers.set("foobar", 42); return true; }
+	 * @example callback(context) { context.modifiers.delete("affinity"); return true; }
+	 */
+	DAMAGE_PIPELINE_PRE_CALCULATE: 'projectfu.pipelines.damage.collect',
```
KSops, line 12: nit: Change the underlying value to more closely match the const name. same with post

KSops (approved):
I didn't check all the nitty gritty stuff but the design looks good to me. Thank you for all the hard work!

Issue #204, opened 2025-01-21
Sheets load extremely slowly with abilities/spells


4m2c4bl3:
I think there needs to be some kind of lazy loading set up for spells, abilities etc in character sheets. 
Right now a level 11 character's sheet is significantly slower to open than one with items removed.
I don't want to imagine how slow they will be once my players hit 20+.

https://github.com/user-attachments/assets/7d303479-f6fe-49f1-b2d0-1f30e780ee68

4m2c4bl3:
I think the fact that _prepareItems seems to iterate over every single item in the context 4 times _while iterating over each item_ might actually be contributing to the problem. Going to try and refactor this function myself

PR #205, opened 2025-01-21
Fix _prepareItems iterating madness

4m2c4bl3:
https://github.com/League-of-Fabulous-Developers/FoundryVTT-Fabula-Ultima/issues/204 

Fixes insane load times on character sheets. For each item in the context we were doing a full for loop for every item in the context again ... four times.


module/sheets/actor-standard-sheet.mjs
```diff
@@ -218,181 +221,158 @@ export class FUStandardActorSheet extends ActorSheet {
...
+			item.defect = item.system.isDefect?.value ? true : false;
+			item.defectMod = item.system.use?.value;
+			item.zeroTrigger = item.system.zeroTrigger?.value;
+			item.zeroEffect = item.system.zeroEffect?.value;
+			item.progressCurr = item.system.progress?.current;
+			item.progressStep = item.system.progress?.step;
+			item.progressMax = item.system.progress?.max;
+
+			if (CLOCK_TYPES.includes(item.type)){
+				const progressArr = [];
```
KSops, line 259: looks like we might have existing utility code for this: generateProgressArray 
https://github.com/League-of-Fabulous-Developers/FoundryVTT-Fabula-Ultima/blob/953945152ff42db884363c60fcbfc9a65b6dcc09/module/documents/items/item.mjs#L647
and
https://github.com/League-of-Fabulous-Developers/FoundryVTT-Fabula-Ultima/blob/dev/module/documents/items/common/progress-data-model.mjs

module/sheets/actor-standard-sheet.mjs
```diff
@@ -218,181 +221,158 @@ export class FUStandardActorSheet extends ActorSheet {
...
-
-					for (let i = 0; i < rp.max; i++) {
-						rpArr.push({
-							id: i + 1,
-							checked: parseInt(rp.current) === i + 1,
-						});
-					}
-
-					item.rpArr = rpArr.reverse();
+			if (RESOURCE_POINT_TYPES.includes(item.type)){
```
KSops, line 269: This might be dead code? I don't see references to `rpArr`

KSops (approved):

KSops:
Nice find!

module/sheets/actor-standard-sheet.mjs
4m2c4bl3: hm, yeah. It uses a different list of valid item types though.
if you make a 'rule' or 'zeroPower' item you seem to be able to see it as a clock, but not edit the details in the attributes tag. I think this should be looked at as a separate bug.

module/sheets/actor-standard-sheet.mjs
4m2c4bl3: I commented it out and none of my character sheets that use resource points seemed affected. The item.system.rp is accessible, so I don't see why modules can't use that?

We could just depreciate it for now, and remove it all together at some good time (when v11 support is dropped maybe)?

Shourn:
The latest changes introduced a merge conflict.
There were two changes to item keys:
- `item.system.mpCost.value` is now `item.system.cost.amount`
- `item.system.target.value` is now `item.system.targeting.rule`

4m2c4bl3:
Done updating item keys

PR #207, opened 2025-01-25
feature: Store pipeline "reverted" state persistently

Shourn:


Azurelol (approved):
Thanks for the follow up!

PR #208, opened 2025-01-25
feat: Action abstraction

Shourn:
- requesting feedback


module/ui/action-test.mjs
```diff
@@ -0,0 +1,52 @@
...
+
+const applyDamageAction = new Action(
+	'applyDamage',
+	(document, element, event) => {
+		console.log(CheckConfiguration.inspect(document).getDamage());
+	},
+	{
+		hasPermission: (document, element) => fromUuidSync(element.dataset.id).testUserPermission(game.user, 'OWNER'),
+	},
+);
```
Azurelol, line 40: When defining an action like so, would it be possible to "chain" the definition of a **revert** action for it? Particularly since the action to revert will usually use the same data.

module/ui/action-test.mjs
```diff
@@ -0,0 +1,52 @@
...
+	someAction.attach(message, html);
+	applyDamageAction.attach(message, html);
+};
+
+export const ActionTest = Object.freeze({
+	initialize: () => {
+		Hooks.on('chatMessage', handleChatCommand);
+		Hooks.on('renderChatMessage', onRenderChatMessage);
+	},
+});
```
Azurelol, line 52: Could it be possible to streamline this in a single function call? Perhaps onto a static function like `Action.register(onChatMessage, actions...)`?

module/ui/action-test.mjs
Shourn: I had given this some thought and decided against it, for now at least.

First, there's not a lot of actions you'd want to revert, really only resource changes and maybe effect application.
Second, reverting doesn't necessarily use the same data.
Honestly I'd be surprised if we found more than 1 or 2 instances where apply and revert data were the same.

As an example, damage application needs to know the target actor, the base damage, the damage type and any special modifiers like "ignores Resistance".
The reverse operation only needs target actor and total damage dealt.
Similarly effect application needs the target actor and the effect definition, but reverting needs the actor and the ID of the created effect.

I feel this is better handled as a separate action.
If you think there's something I missed during my analysis please do let me know.

module/ui/action-test.mjs
Shourn: ~~Should be possible, but would likely need specialized registration methods for different hooks.
Since `Action` is partially defined by using `foundry.abstract.Document` we would require a way to extract the document from the hook.
Since there's only about 4 interesting hooks in this regard and 3 of them share the same signature it shouldn't be a big deal.~~
Thinking about it a bit more `renderChatMessage` would conceivably be the only hook where such a method was of actual use since listeners on sheets usually get attached in `activateListeners`.
A static method `Action.attachAll(document, html, ...actions)` would work though.

Interesting hooks for reference:
- `renderChatMessage` -> document is `ChatMessage` itself
- `renderDocumentSheet`, `renderActorSheet` and `renderItemSheet` -> document can be obtained through `sheet.document`

module/ui/action-test.mjs
Azurelol: No, you are correct and I agree with your assessment. It will not be much more work to add the revert actions.

module/ui/action-test.mjs
Azurelol: Aye, I think that would work nicely.

module/ui/action.mjs
```diff
@@ -0,0 +1,84 @@
...
+		this.#options.requireOwner = options.requireOwner ?? !(typeof options.hasPermission === 'function');
+		this.#options.hasPermission = options.hasPermission;
+	}
+
+	/**
+	 * Attach the action to the document and HTML
+	 * @param {foundry.abstract.Document} document the document represented by the HTML
+	 * @param {jQuery} jQuery the "top level" jQuery object for the HTML representing the document, usually available as a hook param
+	 */
+	attach(document, jQuery) {
```
Azurelol, line 54: I originally had thought to use flags set on message beforehand to quickly exit out of these callbacks but just checking for the item itself may very well be the thing. 

Since there will be dozens of these chat message handlers eventually checking each message, it was a concern for me.

Shourn: I added a `shouldAttach` option that allows canceling listener attachment.
I don't think it will prove necessary for the general case, since I assume DOM traversal through selectors to be a very well optimized operation in browsers and if no matching elements are found no listeners will be attached.
But there might be cases where the name of an action could be reused for different contexts and having a way to stop the listeners from being attached in an "incompatible" context might prove useful.

Azurelol: Yes, I agree.

Shourn:
Was a fine idea, but doesn't really gel with how ApplicationV2 handles action listeners.

PR #210, opened 2025-01-27
Update skill entries

Azurelol:
- Updates many of the skill entries of the compendium to the newer formatting.
- Adds new effects for symbolist class features such as symbol of weakness
- Add skill level to the skill tags


Azurelol:
Tested this one thoroughly.

Issue #206, opened 2025-01-22
Clock types are inconsistant


4m2c4bl3:
FUItem decides an item is a clock in generateProgressArray 
for types ritual and miscAbility

FUStandardActorSheet decides item is a clock in _prepareItems
for types ritual, miscAbility, zeroPower and rule

Items that are zeroPower and rule show up as clocks in the character sheet, but do not have the ability to be edited as clocks in their attributes tag

![Image](https://github.com/user-attachments/assets/28a4b900-9f57-4293-83ca-d65f75af3370)

![Image](https://github.com/user-attachments/assets/147bd38c-a502-473b-85c7-6d991711d796)

PR #211, opened 2025-01-27
Consistent clock types

4m2c4bl3:
https://github.com/League-of-Fabulous-Developers/FoundryVTT-Fabula-Ultima/issues/206


4m2c4bl3:
looking at things, zeroPower doesn't seem to be a main type, it's a sub-type of optionalFeature. so why is it being listed as a type? there's some weirdness going on there, I don't think this is ready to merge. 

also some optional features allow clocks that don't let you set the name ... 
image
😭
i separated this out to try and make these small PRs ...

Azurelol:
Stale.

PR #215, opened 2025-02-02
Update combat system to handle start turn, end turn; improve effect management

Azurelol:
# Combat System, Tracker

Refactors the combat system and tracker to handle two discrete steps between starting a turn and ending it.
Adds hooks for the common events during a combat encounter run by the system:
- Combat Start
- End of Turn
- End of Round
- Combat End

fu_combat_tracker

# Effects Pipeline

Moves the effects class on with the other pipelines.. as it acts like one. It now hooks by default into the combat system in order to manage active effects. When these events are emitted chat messages can be displayed to update users and provide easy buttons to remove effects that have expired.

# Fixes

- Status effects are now cleared properly when temporary effects are cleared.
- Cleans up the deprecated properties used for status effect rendering (Handlebars)


module/helpers/flags.mjs
```diff
@@ -4,7 +4,9 @@
  * @example actor.getFlag(Flags.Scope, Flags.CurrentTurn)
  */
 export const Flags = Object.freeze({
+	CombatantsTurnStarted: 'combatantsTurnStarted',
```
KSops, line 7: might as well align case with other flags?

module/ui/combat-hud.mjs
```diff
@@ -331,16 +343,11 @@ export class CombatHUD extends Application {
...
-		const turnOutOfTurnButtons = html.find('[data-action=take-turn-out-of-turn]');
-		turnOutOfTurnButtons.click((event) => this._doTakeTurnOutOfTurn(event));
-	}
+		html.find('a[data-action=start-turn]').click((event) => ui.combat.handleStartTurn(event));
+		html.find('a[data-action=end-turn]').click((event) => ui.combat.handleEndTurn(event));
 
-	async _doTakeTurn(event) {
-		await ui.combat.handleTakeTurn(event);
+		// TODO: Why was this?
+		// if (!game.user.isGM) return;
```
KSops, line 350: Highlighting for explicit review

module/ui/combat-hud.mjs
```diff
@@ -397,7 +408,8 @@ export class CombatHUD extends Application {
 		Hooks.on('preUpdateScene', this._preUpdateScene.bind(this));
 
 		if (!game.user.isGM) return;
-		game.settings.set(SYSTEM, SETTINGS.optionCombatHudActorOrdering, this._backupOrdering ?? []);
+		// TODO: Enable back?
+		//game.settings.set(SYSTEM, SETTINGS.optionCombatHudActorOrdering, this._backupOrdering ?? []);
```
KSops, line 412: Highlighting for explicit review

module/ui/combat.mjs
```diff
@@ -52,7 +92,8 @@ export class FUCombat extends Combat {
 	 */
 	constructor(data, context) {
 		super(data, context);
-
+		//this._combatant = null;
```
KSops, line 95: leftover?

module/helpers/flags.mjs
Azurelol: Aye, will do.

module/ui/combat.mjs
Azurelol: Ack. Yes it was.

module/ui/combat-hud.mjs
Azurelol: Added back without issue that I saw.

PR #222, opened 2025-02-16
Add combat event hooks

Azurelol:
Add a variety of combat event hooks that can be used by modules (and the system) . These can all be found in `FUHooks`.
- ATTACK_EVENT
- DAMAGE_EVENT 
- COMBAT_EVENT
- GAIN_EVENT
- LOSS_EVENT
- CRISIS_EVENT
- DEFEAT_EVENT
- STATUS_EVENT

## Examples

```js
    Hooks.on('projectfu.events.damage', async event => {
        Azurecompendia.log(`Playing sound effect for damage event: ${event.type} on token: ${event.token.name}`);
        switch (event.type) {
            case "physical":
                playSoundEffect('physical');
                break;
            default:
                break;
        }
    });

    Hooks.on('projectfu.events.gain', async event => {
        Azurecompendia.log(`Playing sound effect for gain event: ${event.resource} on token: ${event.token.name}`);
    });

    Hooks.on('projectfu.events.loss', async event => {
        Azurecompendia.log(`Playing sound effect for loss event: ${event.resource} on token: ${event.token.name}`);
    });

    Hooks.on('projectfu.events.crisis', async event => {
        Azurecompendia.log(`Playing sound effect for crisis event on token: ${event.token.name}`);
    });

    Hooks.on('projectfu.events.combat', async event => {        
        switch (event.type) {
            case 'FU.StartOfCombat':
            case 'FU.EndOfCombat':
                Azurecompendia.log(`Playing sound effect for combat event: ${event.type}`);
                break;

            case 'FU.StartOfTurn':
                Azurecompendia.log(`Playing sound effect for combat event ${event.type} on token ${event.token.name}`);
                break;
            case 'FU.EndOfTurn':
                Azurecompendia.log(`Playing sound effect for combat event ${event.type} on token ${event.token.name}`);
                break;
            default:
                break;
        }
    });
```

### DIFF
--- a/module/checks/check-configuration.mjs
+++ b/module/checks/check-configuration.mjs
@@ -8,6 +8,7 @@ const TARGETS = 'targets';
 const TARGETED_DEFENSE = 'targetedDefense';
 const DIFFICULTY = 'difficulty';
 const DAMAGE = 'damage';
+const TRAITS = 'traits';
 
 /**
  *
@@ -68,6 +69,9 @@ const configure = (check) => {
 };
 
 class CheckConfigurer {
+	/**
+	 * @type {CheckV2, CheckResultV2}
+	 */
 	#check;
 
 	constructor(check) {
@@ -104,12 +108,40 @@ class CheckConfigurer {
 	}
 
 	/**
+	 * @param {String[]|String} traits
+	 * @returns {CheckConfigurer}
+	 */
+	addTraits(...traits) {
+		if (!this.#check.additionalData[TRAITS]) {
+			this.#check.additionalData[TRAITS] = [];
+		}
+		traits.forEach((t) => this.#check.additionalData[TRAITS].push(t.toLowerCase()));
+		return this;
+	}
+
+	/**
 	 * @param {FUItem} item
 	 * @param {FUActor} actor
 	 * @return {CheckConfigurer}
 	 */
 	addItemAccuracyBonuses(item, actor) {
 		return this.addModelAccuracyBonuses(item.system, actor);
+	}
+
+	/**
+	 * @description Add the common traits of a weapon
+	 * @param {WeaponDataModel} system
+	 */
+	addWeaponTraits(system) {
+		return this.addTraits(system.category.value, system.type.value, system.hands.value);
+	}
+
+	/**
+	 * @description Add the common traits of an NPC attack
+	 * @param {BasicItemDataModel} system
+	 */
+	addAttackTraits(system) {
+		return this.addTraits(system.type.value, system.damageType.value);
 	}
 
 	/**
@@ -367,8 +399,12 @@ const inspect = (check) => {
 
 /**
  * @description Given a {@link CheckResultV2} object, provides additional information from it
+ * @remarks Provides read-only access, to be used after {@linkcode CheckConfigurer}
  */
 class CheckInspector {
+	/**
+	 * @type CheckResultV2
+	 */
 	#check;
 
 	constructor(check) {
@@ -411,6 +447,13 @@ class CheckInspector {
 	}
 
 	/**
+	 * @return {String[]|null}
+	 */
+	getTraits() {
+		return this.#check.additionalData[TRAITS] ?? null;
+	}
+
+	/**
 	 * @return {TargetData[]|null}
 	 */
 	getTargets() {
@@ -422,6 +465,20 @@ class CheckInspector {
 	 */
 	getTargetsOrDefault() {
 		return this.getTargets() || [];
+	}
+
+	/**
+	 * @returns {Boolean}
+	 */
+	isCritical() {
+		return this.getCheck().critical;
+	}
+
+	/**
+	 * @returns {Boolean}
+	 */
+	isFumble() {
+		return this.getCheck().fumble;
 	}
 
 	/**
@@ -475,6 +532,7 @@ class CheckInspector {
 					total: damage.total,
 					type: damage.type,
 					extra: damage.extra,
+					traits: this.getTraits(),
 				},
 				translation: {
 					damageTypes: FU.damageTypes,
--- a/module/checks/check-configuration.mjs
+++ b/module/checks/check-configuration.mjs
@@ -447,14 +447,14 @@ class CheckInspector {
 	}
 
 	/**
-	 * @return {String[]|null}
+	 * @return {String[]}
 	 */
 	getTraits() {
 		return this.#check.additionalData[TRAITS] ?? null;
 	}
 
 	/**
-	 * @return {TargetData[]|null}
+	 * @return {TargetData[]}
 	 */
 	getTargets() {
 		return this.#check.additionalData[TARGETS] ? foundry.utils.duplicate(this.#check.additionalData[TARGETS]) : null;
--- a/module/checks/check-configuration.mjs
+++ b/module/checks/check-configuration.mjs
@@ -450,7 +450,7 @@ class CheckInspector {
 	 * @return {String[]}
 	 */
 	getTraits() {
-		return this.#check.additionalData[TRAITS] ?? null;
+		return this.#check.additionalData[TRAITS] ?? [];
 	}
 
 	/**
--- a/module/checks/common-events.mjs
+++ b/module/checks/common-events.mjs
@@ -1,0 +1,263 @@
+import { FUHooks } from '../hooks.mjs';
+import { Targeting } from '../helpers/targeting.mjs';
+
+/**
+ * @description Contains both information about a target in a combat event
+ * @typedef EventTarget
+ * @property {Token} token
+ * @property {FUActor} actor
+ * @property {TargetData|null} data
+ */
+
+/**
+ * @typedef ItemReference
+ * @property {String} name
+ * @property {String} fuid
+ */
+
+/**
+ * @param {FUItem} item
+ * @returns {ItemReference}
+ */
+function toItemReference(item) {
+	return {
+		name: item.name,
+		fuid: item.system.fuid,
+	};
+}
+
+/**
+ * @description Dispatched when an actor makes an attack (skill/spell)
+ * @typedef AttackEvent
+ * @property {ItemReference} item
+ * @property {FU.damageTypes} type
+ * @property {Set<String>} traits
+ * @property {FUActor} actor
+ * @property {Token} token
+ * @property {EventTarget[]} targets
+ */
+
+/**
+ * @description Dispatches an event to signal an attack event
+ * @param {CheckInspector} inspector
+ * @param {FUActor} actor
+ * @param {FUItem} item
+ */
+function attack(inspector, actor, item) {
+	const traits = inspector.getTraits();
+	const targets = inspector.getTargets();
+	const eventTargets = getEventTargets(targets);
+
+	/** @type AttackEvent  **/
+	const event = {
+		item: toItemReference(item),
+		actor: actor,
+		type: inspector.getDamage()?.type,
+		token: actor.resolveToken(),
+		targets: eventTargets,
+		traits: new Set(traits),
+	};
+	Hooks.call(FUHooks.ATTACK_EVENT, event);
+}
+
+/**
+ * @description Dispatched when an actor suffers damage
+ * @typedef DamageEvent
+ * @property {FU.damageTypes} type
+ * @property {Number} amount
+ * @property {FUActor} actor
+ * @property {Token} token
+ * @property {EventTarget} source
+ * @property {Set<String>} traits
+ */
+
+/**
+ * @description Dispatches an event to signal damage taken by an actor
+ * @param {FU.damageTypes} type
+ * @param {Number} amount
+ * @param {Set<String>} traits
+ * @param {FUActor} actor
+ * @param {FUActor} sourceActor
+ */
+function damage(type, amount, traits, actor, sourceActor) {
+	const source = getEventTargetFromActor(sourceActor);
+	/** @type DamageEvent  **/
+	const damageEvent = {
+		amount: amount,
+		type: type,
+		actor: actor,
+		token: actor.resolveToken(),
+		source: source,
+		traits: traits,
+	};
+	Hooks.call(FUHooks.DAMAGE_EVENT, damageEvent);
+}
+
+/**
+ * @description Dispatched when an actor enters crisis
+ * @typedef CrisisEvent
+ * @property {FUActor} actor
+ */
+
+/**
+ * @description Dispatched when an actor is reduced to 0 HP
+ * @typedef DefeatEvent
+ * @property {FUActor} actor
+ */
+
+/**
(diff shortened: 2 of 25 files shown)


module/checks/check-configuration.mjs
```diff
@@ -7,6 +7,7 @@ const TARGETS = 'targets';
 const TARGETED_DEFENSE = 'targetedDefense';
 const DIFFICULTY = 'difficulty';
 const DAMAGE = 'damage';
+const TRAITS = 'trails';
```
KSops, line 10: typo

Azurelol: Ack. Thanks!

module/documents/actors/actor.mjs
```diff
@@ -13,6 +13,19 @@ import { Flags } from '../../helpers/flags.mjs';
  * @property {Boolean} inCombat
  * @property {String} id The canonical identifier for this Document.
  * @property {String} uuid A Universally Unique Identifier (uuid) for this Document instance.
+ * @property {Function<[]>} getActiveTokens Retrieve an Array of active tokens which represent this Actor in the current
```
KSops, line 16: looks like this method takes a optional bool param?

module/hooks.mjs
```diff
@@ -42,6 +42,95 @@ export const FUHooks = {
 	ROLL_STUDY: 'studyRoll',
 	/**
 	 * @description Invoked when there's a change in the combat state
+	 * @example callback(event)
+	 * @remarks Uses {@link CombatEvent}
 	 */
-	COMBAT_EVENT: 'projectfu.combat',
+	COMBAT_EVENT: 'projectfu.events.combat',
```
KSops, line 48: This is a breaking change right? Have proper stakeholders been notified?

module/hooks.mjs
```diff
@@ -42,6 +42,95 @@ export const FUHooks = {
...
+	 */
+	GAIN_EVENT: 'projectfu.events.gain',
+	/**
+	 * @description Invoked after resource loss has been applied to an actor
+	 * @example callback(event)
+	 * @remarks Uses {@link LossEvent}
+	 */
+	LOSS_EVENT: 'projectfu.events.loss',
+	/**
+	 * @description Invoked after an actor enters or exits crisis
```
KSops, line 68: Should explain the relationship/event sequence for when a damage packet will trigger both crisis and KO

module/pipelines/damage-pipeline.mjs
```diff
@@ -378,6 +378,20 @@ async function process(request) {
 		const damageTaken = -context.result;
 		updates.push(actor.modifyTokenAttribute('resources.hp', damageTaken, true));
 		actor.showFloatyText(`${damageTaken} HP`, `red`);
+
+		// Dispatch event
+		/** @type DamageEvent  **/
```
KSops, line 383: On my local env I wasnt able to see the typedef properly linked. Does it work for you?
If not I think this line will help with typedef import at the top(ish) of the file:
`
/** @typedef {import('../hooks.mjs').DamageEvent DamageEvent} */
`
and then for the type here:
`
/** @type {DamageEvent}  **/
`

module/pipelines/effects.mjs
```diff
@@ -90,8 +90,10 @@ export async function onManageActiveEffect(event, owner) {
 			return createTemporaryEffect(owner, listItem.dataset.effectType);
 		case 'edit':
 			return resolveEffect().sheet.render(true);
-		case 'delete':
-			return resolveEffect().delete();
+		case 'delete': {
+			const effect = resolveEffect();
+			return effect.delete();
+		}
```
KSops, line 96: leftover from debugging?

module/hooks.mjs
```diff
@@ -42,6 +42,95 @@ export const FUHooks = {
...
+	 */
+	CRISIS_EVENT: 'projectfu.events.crisis',
+	/**
+	 * @description Invoked after an actor is reduced to 0 hit points
+	 * @example callback(event)
+	 * @remarks Uses {@link DefeatEvent}
+	 */
+	DEFEAT_EVENT: 'projectfu.events.defeat',
+	/**
+	 * @description Dispatched after an actor has a status effect applied on them.
```
KSops, line 80: It looks like this is also for when it's removed. We should also document the timing of the event. e.g. will the effect exist before/after the event? (I agree with the way you did the timing on the events)

module/ui/combat.mjs
```diff
@@ -12,7 +12,10 @@ export const HOSTILE = 'hostile';
  * @description Dispatched by the combat during its lifetime
  * @property {FU.combatEvent} type The type of event
  * @property {Number} round The round the event is taking place in
+ * @property {Combatant} combatant The current combatant taking a turn, which can be null.
  * @property {FUActor|*} actor The actor involved in the event, which can be null.
+ * @property {Token|*} token The token of the combatant taking a turn, which can be null.
+ * @property {Combatant[]} combatants The actors involved in the combat
```
KSops, line 18: It seems that some of the fields are conditionally populated based on the event type. Documentation might be good to set expectations as to what fields are present vs undefined/null when

KSops:
As a general comment, I'm not sure the JSDoc typedefs are working, but that could also be my VSCode acting up.

Azurelol:
> As a general comment, I'm not sure the JSDoc typedefs are working, but that could also be my VSCode acting up.

Oh I definitely had issues referencing types in VScode; hence my switch to Webstorm.

module/documents/actors/actor.mjs
Azurelol: Aye; it's a pain to document functions with JSDOC; will add.

module/hooks.mjs
Azurelol: It was introduced in the last alpha, hasn't been used by anyone yet. It and the other events like it are mostly meant for SFX.

module/hooks.mjs
Azurelol: Sure thing.

module/pipelines/effects.mjs
Azurelol: Ack. Thank you.

module/hooks.mjs
Azurelol: Will do.

module/ui/combat.mjs
Azurelol: Good point.

module/checks/check-configuration.mjs
```diff
@@ -112,6 +128,22 @@ class CheckConfigurer {
 		return this.addModelAccuracyBonuses(item.system, actor);
 	}
 
+	/**
+	 * @description Add the common traits of a weapon
+	 * @param {WeaponDataModel} system
```
KSops, line 133: It's actually the `system` property of `WeaponDataModel` right? If there's no good type for it, I'd rather just leave it as any or object or whatever denotes an unknown type. Better to lack info than be misleading.

module/checks/check-configuration.mjs
```diff
@@ -102,6 +106,18 @@ class CheckConfigurer {
...
 
+	/**
+	 * @param {String[]} traits
+	 * @returns {CheckConfigurer}
+	 */
+	addTraits(...traits) {
+		if (!this.#check.additionalData[TRAITS]) {
+			this.#check.additionalData[TRAITS] = [];
+		}
+		traits.forEach((t) => this.#check.additionalData[TRAITS].push(t));
```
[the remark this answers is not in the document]
KSops: ah looks like an alternative is
`this.#check.additionalData[TRAITS].push(...traits);`

module/checks/common-events.mjs
```diff
@@ -0,0 +1,263 @@
+import { FUHooks } from '../hooks.mjs';
+import { Targeting } from '../helpers/targeting.mjs';
+
+/**
+ * @description Contains both information about a target in a combat event
```
KSops, line 5: Says "both information" but there's multiple properties?

module/checks/common-events.mjs
```diff
@@ -0,0 +1,263 @@
...
+
+/**
+ * @description Dispatches an event to signal an attack event
+ * @param {CheckInspector} inspector
+ * @param {FUActor} actor
+ * @param {FUItem} item
+ */
+function attack(inspector, actor, item) {
+	const traits = inspector.getTraits();
+	const targets = inspector.getTargets();
```
KSops, line 48: I think getTraits and getTargets can return null. In these cases it looks like there might be exceptions thrown?

module/checks/check-configuration.mjs
KSops: oh nvm that is the actual type. my mistake

module/checks/common-events.mjs
```diff
@@ -0,0 +1,263 @@
...
+}
+
+/**
+ * @description Dispatched when an actor suffers damage
+ * @typedef DamageEvent
+ * @property {FU.damageTypes} type
+ * @property {Number} amount
+ * @property {FUActor} actor
+ * @property {Token} token
+ * @property {EventTarget} source
```
KSops, line 70: EventTarget | null

module/hooks.mjs
```diff
@@ -42,6 +42,56 @@ export const FUHooks = {
...
+	 */
+	LOSS_EVENT: 'projectfu.events.loss',
+	/**
+	 * @description Dispatched after an actor enters or exits crisis.
+	 * @example callback(event)
+	 * @remarks Uses {@link CrisisEvent}. This can happen after a {@link DAMAGE_EVENT}.
+	 */
+	CRISIS_EVENT: 'projectfu.events.crisis',
+	/**
+	 * @description Invoked after an actor is reduced to 0 hit points
```
KSops, line 86: I guess it's an edge case, but what happens if an actor gains HP? Does it lose these statuses and also trigger these events?
Also, we should outline that this always comes after crisis

module/hooks.mjs
```diff
@@ -42,6 +42,56 @@ export const FUHooks = {
...
+	 * @example callback(event)
+	 * @remarks Uses {@link DefeatEvent}. This can happen after a {@link DAMAGE_EVENT}.
+	 */
+	DEFEAT_EVENT: 'projectfu.events.defeat',
+	/**
+	 * @description Dispatched after an actor has a status effect applied on them.
+	 * @example callback(event)
+	 * @remarks Uses {@link StatusEvent}. It happens AFTER the status effect has been applied.
```
KSops, line 94: After applied and before removed.

module/ui/combat.mjs
KSops: If nullable, we can probably express union with null
`type | null`

module/pipelines/damage-pipeline.mjs
```diff
@@ -519,11 +523,13 @@ function onRenderChatMessage(message, jQuery) {
  * @param {InlineSourceInfo} sourceInfo
  * @param {import('../helpers/typedefs.mjs').BaseDamageInfo} baseDamageInfo
  * @param {import('./damage-customizer.mjs').ExtraDamageInfo} extraDamageInfo
+ * @param {String[]} traits
```
KSops, line 526: `| null` (and add a guard for null traits)

src/packs/skills/classFeature_Solemn_Tone_yidpNKJQMA7dSBsc.json
```diff
@@ -13,7 +13,7 @@
     },
     "featureType": "projectfu.tone",
     "data": {
-      "description": "<p>Each target recovers from<strong>【@DATA[key.status]】</strong>. Each target also gains <strong>【@DATA[key.resistance]】</strong>until the <strong>start of your next turn</strong>.</p>"
+      "description": "<p>Each target recovers from<strong>【@DATA[key.statusLocal]】</strong>and also gains <strong>【@AFFINITY[@DATA[key.type] resistance]】</strong>until the <strong>start of your next turn</strong>.</p>"
```
KSops, line 16: Some of these descriptions are using `<strong>` on the `[` brackets and some are not. Just checking that we're following consistent styling in our descriptions.

module/documents/items/classFeature/chanter/verses-application.mjs
```diff
@@ -35,15 +33,14 @@ async function getDescription(model, useAttributes = false) {
...
 		rollData.key = {
-			type: game.i18n.localize(FU.damageTypes[keyData.type]),
-			damage: `@DMG[${SCALING_FORMULA} ${keyData.type}]`,
-			status: `@EFFECT[${keyData.status}]`,
-			attribute: game.i18n.localize(FU.attributeAbbreviations[keyData.attribute]),
-			boost: `@EFFECT[${keyData.attribute}-up]`,
-			resistance: `@AFFINITY[${keyData.type} resistance]`,
-			normal: `@AFFINITY[${keyData.type} none]`,
-			vulnerability: `@AFFINITY[${keyData.type} vulnerability]`,
-			recovery: `@GAIN[${SCALING_FORMULA} ${keyData.recovery}]`,
```
KSops, line 46: The removed fields were unused? Is it possible modules/macros were using them?

module/checks/common-events.mjs
Azurelol: Whoops.

module/checks/common-events.mjs
```diff
@@ -0,0 +1,263 @@
...
+ * @property {EventTarget[]} targets
+ */
+
+/**
+ * @description Dispatches an event to signal an attack event
+ * @param {CheckInspector} inspector
+ * @param {FUActor} actor
+ * @param {FUItem} item
+ */
+function attack(inspector, actor, item) {
```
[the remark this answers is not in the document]
Azurelol: They do have descriptions in the comments regarding that; I wished to keep the convention established in `CommonSections`.

module/checks/common-events.mjs
Azurelol: A while ago they would, but now they return empty arrays if nothing instead. I updated the comments.

module/checks/common-events.mjs
Azurelol: Yes.

module/hooks.mjs
Azurelol: It loses the statuses, does not trigger events again when going above the threshold.

module/hooks.mjs
Azurelol: Aye.

module/pipelines/damage-pipeline.mjs
Azurelol: As mentioned above, this should never be null. Empty on the worst case.

module/checks/common-events.mjs
KSops: getTargets returns an empty array, but getTraits still returns null doesnt it?

Azurelol: I see it.

KSops (approved):

PR #224, opened 2025-02-19
fix: update silk shirt mdef to correct value of 2

CGavinMullis:
In the core rulebook, the silk shirt gives INS die + 2 M. Defense

image

However in the foundry module the value is incorrectly INS die + 1 M. Defense

image


CGavinMullis:
I couldn't find anything in the errata that might indicate this was purposeful

spyrella:
My bad, it's a user error on my part

PR #229, opened 2025-02-20
Add two distinct evaluations for the damage data model

Azurelol:
It was brought to my attention that we actually need two evaluations when adding extra damage from expressions:
- As a bonus when making the roll for the skill (Shadow Strike)
- As a bonus when applying the damage on the target (Cheap Shot)

I updated the fields on the data model and adjusted things accordingly.


Azurelol:
Don't panic! The culprit behind the big line count change came from the formatter redoing the line endings on `en.json`. They should be LF now.

module/documents/actors/common/overrides-data-model.mjs
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * @property {String} damageType The damage type of all actions performed by the character
+ * @property {Number} turns Number of additional turns this character can take each round
+ */
+export class OverridesDataModel extends foundry.abstract.DataModel {
+	static defineSchema() {
+		const { StringField, NumberField } = foundry.data.fields;
+		return {
+			turns: new NumberField({ initial: 1, min: 1, integer: true, nullable: false }),
```
Shourn, line 11: Not a fan of having this under `overrides`.
Personally, I feel like having this under `bonuses` and starting with `0` is more intuitive.
This would make the usual usecases of "lose a turn"/"gain an additional turn" very straightforward and applicable to both PCs and NPCs: \["system.bonuses.turns", "add", "1"].

module/helpers/inline-affinity.mjs
```diff
@@ -96,23 +96,37 @@ async function onDropActor(actor, sheet, { type, sourceInfo, affinity, value, ig
...
-	const attributeKey = `system.affinities.${type}.current`;
-	/** @type Number **/
+	// If overriding the character's damage type for all actions
+	let attributeKey;
+	let attributeValue;
+	let name;
 	const localizedType = game.i18n.localize(FU.damageTypes[type]);
-	const affValue = FU.affValue[value];
-	const localizedValue = game.i18n.localize(FU.affType[affValue]);
+	if (value === 'damage') {
```
Shourn, line 104: I feel like this is conflating different usecases by lumping together changes to affinities and changes to outgoing damage.
It's not a 100% fit, but the inline weapon enchantments feel like a more sensible place for this.

src/packs/skills/skill_Cheap_Shot_c2XRt5pvOab25Tbe.json
```diff
@@ -48,7 +48,8 @@
       "hasDamage": true,
       "value": 0,
       "type": "",
-      "extra": "$sl+$tsc"
+      "onRoll": "",
+      "onApply": "$sl+$tsc"
```
Shourn, line 52: This formula is actually not correct.
"Cheap Shot" only deals its bonus damage if the target is suffering from at least 1 status effect.
Untested, but `(min($tsc, 1) * $sl) + $tsc` might work.

Shourn (changes requested):

module/documents/actors/common/overrides-data-model.mjs
```diff
@@ -0,0 +1,15 @@
...
+			damageType: new StringField({ initial: '', choices: Object.keys(FU.damageTypes), blank: true, nullable: false }),
```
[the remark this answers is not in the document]
Azurelol: Very well. I think we could make sure of a nice table  on a feature request that lists the use cases. I will sub-divide the damage override data structure.

module/documents/items/common/damage-data-model-v2.mjs
```diff
@@ -14,7 +15,8 @@ export class DamageDataModelV2 extends foundry.abstract.DataModel {
 			hasDamage: new BooleanField(),
 			hrZero: new BooleanField(),
 			value: new NumberField({ initial: 0, integer: true, nullable: false }),
-			extra: new StringField({ nullable: true }),
+			onRoll: new StringField({ nullable: true }),
```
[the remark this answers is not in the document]
Azurelol: Ah I wasn't aware of this option. Will do.

module/documents/actors/common/overrides-data-model.mjs
Azurelol: Oh that's a great idea.

module/helpers/inline-affinity.mjs
Azurelol: I wanted to avoid having to define a new inline command, but I may as well at this point.

src/packs/skills/skill_Cheap_Shot_c2XRt5pvOab25Tbe.json
Azurelol: Ah. Will try it.

Azurelol: Yup! 
image

Shourn (approved):

Issue #236, opened 2025-03-06
Optimize hooks that trigger a Combat HUD refresh


Element-Re:
Currently, the Combat HUD registers a number of event hooks that trigger a re-render of the Combat HUD. Unfortunately, these hooks always trigger a re-render regardless of the circumstances of the event. For example, if a token moves in the scene, that would trigger a re-render, despite the fact that this update has nothing to do with the data displayed in the tracker.

Considering that re-rendering the combat HUD is a non-trivial task as far as browser resources go, it would be better to optimize the event handlers to be smarter about when re-renders are triggered.

Element-Re:
I plan to tackle this issue myself.

PR #241, opened 2025-03-06
fix: optimize combat hud hooks

Element-Re:
Issue #236


KSops:
Dont forget to target dev

PR #245, opened 2025-03-08
Combat turn V3

Azurelol:
- Standardizes the combat turn button across the combat hud and tracker
- Fixes taking turns out of order
- Show info for other combatants when someone's taking a turn
- Add item event

image

### DIFF
--- a/module/ui/combat-tracker.mjs
+++ b/module/ui/combat-tracker.mjs
@@ -23,7 +23,6 @@ export class FUCombatTracker extends CombatTracker {
 		/** @type FUCombat **/
 		const combat = data.combat;
 		if (combat) {
-			// We assign it afterwards, since factions will be assigned if empty
 			combat.populateData(data);
 			// We add more data to the turns objects
 			data.turns = data.turns?.map((turn) => {
@@ -106,8 +105,9 @@ export class FUCombatTracker extends CombatTracker {
 		return turns.reduce(
 			(agg, combatantData) => {
 				const combatant = combat.combatants.get(combatantData.id);
-				// The tracker rendering needs this!
+				// The tracker rendering needs this! Do not remove!
 				combatantData.faction = combatant.faction;
+				combatantData.isOwner = combatant.isOwner;
 				if (combatant.token.disposition === foundry.CONST.TOKEN_DISPOSITIONS.FRIENDLY) {
 					agg.friendly.push(combatantData);
 				} else {
--- a/module/ui/combat-tracker.mjs
+++ b/module/ui/combat-tracker.mjs
@@ -23,18 +23,9 @@ export class FUCombatTracker extends CombatTracker {
 		/** @type FUCombat **/
 		const combat = data.combat;
 		if (combat) {
-			data.turnStarted = combat.isTurnStarted;
-			// The current combatant, if any
-			data.combatant = combat.combatant;
-			// What faction's turn it is
-			data.currentTurn = combat.getCurrentTurn();
-			// ID : Turns Left
-			data.turnsLeft = this.countTurnsLeft(combat);
-			// ID : Total Turns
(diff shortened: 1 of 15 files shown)


Azurelol:
Don't be afraid! The line changes come from fixing the line endings for the package files.

module/ui/combat.mjs
```diff
@@ -345,6 +355,68 @@ export class FUCombat extends Combat {
...
+			console.debug(`- Combatant: ${combatant.name}, faction: ${combatant.faction}, isOwner: ${combatant.isOwner}`);
+		}
+	}
+
+	/**
+	 * @param turns
+	 * @param {FUCombat} combat
+	 * @return {Object.<"friendly"|"neutral"|"hostile", {}[]>}
+	 */
+	getFactions() {
```
Element-Re, line 390: The original version of this method was not cleaned up combat-tracker.mjs: https://github.com/League-of-Fabulous-Developers/FoundryVTT-Fabula-Ultima/blob/a546f591a8a17b4de33be131ba8d69a628549a9b/module/ui/combat-tracker.mjs#L103

module/ui/combat.mjs
```diff
@@ -345,6 +355,68 @@ export class FUCombat extends Combat {
...
+		data.totalTurns = this.combatants.reduce((agg, combatant) => {
+			agg[combatant.id] = combatant.totalTurns;
+			return agg;
+		}, {});
+		data.turnStarted = this.isTurnStarted;
+		// The current combatant, if any
+		data.combatant = this.combatant;
+		// ID : Turns Left
+		data.turnsLeft = this.countTurnsLeft();
+		data.factions = this.getFactions();
```
Element-Re, line 372: Might need to pass `data` into this (see below)

module/ui/combat.mjs
```diff
@@ -345,6 +355,68 @@ export class FUCombat extends Combat {
...
+		}
+	}
+
+	/**
+	 * @param turns
+	 * @param {FUCombat} combat
+	 * @return {Object.<"friendly"|"neutral"|"hostile", {}[]>}
+	 */
+	getFactions() {
+		return this.combatants.reduce(
```
Element-Re, line 391: This is the crux of the issue with with the missing animated token thumbnails and other elements like effects or resources. The previous version of getFactions had data.turns passed in from populateData, and it was these turns that was iterated over in the combat tracker handlebars template. This used the combatants collection on the combat, which is missing information compared to the turns array previously passed in (such as a dynamic url based on either the token's image itself, or a base64 png display url generated from it instead.

module/ui/combat.mjs
Azurelol: Aye.

module/ui/combat.mjs
Azurelol: This was correct, as the combat tracker needs different data.

module/ui/combat-tracker.mjs
```diff
@@ -23,18 +23,9 @@ export class FUCombatTracker extends CombatTracker {
...
-			// What faction's turn it is
-			data.currentTurn = combat.getCurrentTurn();
-			// ID : Turns Left
-			data.turnsLeft = this.countTurnsLeft(combat);
-			// ID : Total Turns
-			data.totalTurns = combat.combatants.reduce((agg, combatant) => {
-				agg[combatant.id] = combatant.totalTurns;
-				return agg;
-			}, {});
+			// We assign it afterwards, since factions will be assigned if empty
```
KSops, line 26: I'm not sure I understand this. What do we mean by afterwards? The logic is not deferred here right?

KSops (approved):

module/ui/combat-tracker.mjs
Azurelol: Whops, forgot to remove this comment. (Old implementation)

PR #246, opened 2025-03-09
Fix combat hud bottom button drag

KSops:
- fixes a bug in drag/drop where UI would jump off-screen
- fixes dragging issue with the bottom button where starting the drag would cause the top of the UI to jump to the cursor position rather than staying where the drag started
- prettier changes applied to code automatically; no logic changes after ~line 700

### DIFF
--- a/module/ui/combat-hud.mjs
+++ b/module/ui/combat-hud.mjs
@@ -23,6 +23,15 @@ export class CombatHUD extends Application {
 		this._emptyImage = document.createElement('img');
 		this._emptyImage.src = 'data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7';
 
+		// Drag trackers
+		this.dragInitialX = 0;
+		this.dragInitialY = 0;
+		this.dragInitialTop = 0;
+		this.dragInitialLeft = 0;
+
+		// TODO: Move such browser checks to a higher scope
+		this.isFirefox = navigator.userAgent.toLowerCase().includes('firefox');
+
 		console.debug(`Combat HUD: Constructing`);
 		Hooks.callAll('combatHudInit', this);
 		this.#hooks.push({ hook: 'createCombatant', func: this._onUpdateHUD.bind(this) });
@@ -193,10 +202,8 @@ export class CombatHUD extends Application {
 				actor: combatant.actor,
 				token: combatant.token,
 				effects: activeEffects,
-				img: game.settings.get(SYSTEM, SETTINGS.optionCombatHudPortrait) === 'token' ? 
-					// token._source should contain the most current version of the token's texture.
-					combatant.token._source.texture.src : 
-					combatant.actor.img,
+				// token._source should contain the most current version of the token's texture.
+				img: game.settings.get(SYSTEM, SETTINGS.optionCombatHudPortrait) === 'token' ? combatant.token._source.texture.src : combatant.actor.img,
 				trackedResourcePart1: trackedResourcePart1,
 				trackedResourcePart2: trackedResourcePart2,
 				trackedResourcePart3: trackedResourcePart3,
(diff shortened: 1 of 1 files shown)


module/ui/combat-hud.mjs
```diff
@@ -193,10 +202,8 @@ export class CombatHUD extends Application {
 				actor: combatant.actor,
 				token: combatant.token,
 				effects: activeEffects,
-				img: game.settings.get(SYSTEM, SETTINGS.optionCombatHudPortrait) === 'token' ? 
-					// token._source should contain the most current version of the token's texture.
-					combatant.token._source.texture.src : 
-					combatant.actor.img,
+				// token._source should contain the most current version of the token's texture.
+				img: game.settings.get(SYSTEM, SETTINGS.optionCombatHudPortrait) === 'token' ? combatant.token._source.texture.src : combatant.actor.img,
```
KSops, line 206: I do like the original formatting but prettier says otherwise

module/ui/combat-hud.mjs
```diff
@@ -356,32 +363,60 @@ export class CombatHUD extends Application {
 
 	_doHudDragStart(event) {
 		event.originalEvent.dataTransfer.setDragImage(this._emptyImage, 0, 0);
+
+		// Set drag tracking vars
+		const nativeEvent = event.originalEvent;
+		const elementPos = this.element.position();
```
KSops, line 369: stored as a var in case there's overhead in obtaining `position()` twice

module/ui/combat-hud.mjs
```diff
@@ -356,32 +363,60 @@ export class CombatHUD extends Application {
...
+			cancelAnimationFrame(this._dragAnimationFrame);
+		}
 		this._dragAnimationFrame = requestAnimationFrame(() => {
-			this.element.css('left', this._dragOffsetX + dragPosition.x);
-			this.element.css('top', this._dragOffsetY + dragPosition.y);
+			// Calculate deltas
+			const deltaX = x - this.dragInitialX;
+			const deltaY = y - this.dragInitialY;
+
+			// Calculate and log final values before applying
```
KSops, line 413: there's no logging done here. will fix comment

Azurelol (approved):

Azurelol:
Looks good. Confirm once it's ready to merge, as I read you want to fix a comment?

KSops:
I'd like to wait on Element to review since it's related to his changes

module/ui/combat-hud.mjs
```diff
@@ -356,32 +363,60 @@ export class CombatHUD extends Application {
...
-		// during drag events. A workaround is to bind a seperate handler
-		// to the dragover event instead, which gets processed directly
-		// before any drag event and allows us to record the current drag position.
-		this._dragX = event.clientX;
-		this._dragY = event.clientY;
+		// Keep this check; drag events can trigger with (0,0) when outside the window or target
+		// and they should be treated as invalid
+		if (event.clientX <= 0 || event.clientY <= 0) return;
+
+		this.updatePosition(event.clientX, event.clientY);
```
Element-Re, line 385: I don't think we're supposed to update the position here. This is just to store the position of clientX, clientY so we can use them in the _doHudDrag handler. In fact, this handler gets called when *any* element is being dragged (that's why it's a workaround) but as long as it's just updating a couple of variables the impact should be minimal.

module/ui/combat-hud.mjs
```diff
@@ -356,32 +363,60 @@ export class CombatHUD extends Application {
...
 	}
 
 	_doHudDrag(event) {
-		let dragPosition;
+		// Keep this check; drag events can trigger with (0,0) when outside the window or target
+		// and they should be treated as invalid
+		if (event.clientX <= 0 || event.clientY <= 0) return;
 
-		if (navigator.userAgent.toLowerCase().includes('firefox')) {
-			// Workaround: FireFox doesn't populate event.<clientX/clientY> during drag events
```
Element-Re, line 374: Same issue as [here](https://github.com/League-of-Fabulous-Developers/FoundryVTT-Fabula-Ultima/pull/246/files#r1986432382). We really do need to perform this check, or maybe just check if _dragX _dragY are set. If we update the position in the other handler we'll see the hud being dragged around when any element is being dragged I'm pretty sure.

Element-Re:
Beyond my comments about updating the position in the firefox workaround handler, I don't see any issues.

PR #251, opened 2025-03-13
Overhaul active effect system

Azurelol:
Overhauls the active effect model for the system:

- Implements a data model, under `system`.
- Migrates existing properties that were serialized in flags to this newer data model.
- Provides a duration property group, which can be managed during combat. These include `event` (start of turn, end of turn, ...), `interval` (integer), `tracking` (self or source).
- Add support for referencing the source of the application in expressions, which proves invaluable for the automation of certain effects such as `Breach`, or `Cover`.
- Adds optional labeling to inline commands, through the  `@COMMAND[...]{label}`  syntax.
- Implements a system for tracking active effects during combat, through the use of previously implemented events.
- Adds options for the effect management system, in the system options.
- Adds automation to the core set of skills and actions, with more to come.

fu_guard_automation


Azurelol:
I am doing the finishing touches, but the PR is mostly ready for review.

PR #270, opened 2025-04-03
Add inline clock, clock expressions

Azurelol:
- Add inline clock commands, `@CLOCK[id update step]`, `@CLOCK[id reset]`
- Add expression macro `&cs`, to evaluate the current sections in a clock by the id
- Add automation for many of the esper entries


Azurelol:
> I could mention that system.rp functions similarly to clocks, and can also benefit from it's own inline command

Oh? What skills use these resource points?

spyrella:
> Oh? What skills use these resource points?

Here's a comprehensive list
Skills
Necromancer
Rondo of Nightmare (Grave Points): may spend 2 Grave Points
Beyond The Realm of the Dead (Grave Points): gain 1 Grave Points
Fear is the Key (Grave Points): may spend 1 Grave Points

Merchant
I've Heard of It! (Trade Points): spend 1 Trade Point 
Private Stock (Trade Points): spend 1 Trade Point
Real Treasure (Trade Points): gain SL+1 Trade Points
Wind of Trade (Trade Points): gain SL+1 Trade Points, spend 2 Trade Points

Class Feature
Arcanum Rework
Banshee [Whirlwind Points] / Lilith [Gnosis Points]

Heroic Skill
Rising Tide (Bravery Points): gain 1 Bravery Point, must spend all  end of scene
Swirling Swarm (Momentum Points): gain 1 Momentum Point, lose all end of scene
Nebulization (Nebulization Points): gain 1 Nebulization Point or 2 Nebulization Points if potion created vvia alchemy/gadget, lose all at end of scene

Heroic Style
Seven Weapons Style [Sequence Points]

Azurelol:
> I could mention that system.rp functions similarly to clocks, and can also benefit from it's own inline command

I think we can reuse the same command for it, particularly since we can use custom labels.

PR #273, opened 2025-04-07
Add progress data to active effect model

Azurelol:
## Description

Adds progress data to the active effect model, allowing us to embed progress tracks (to serve as either clocks or resource points) into any kind of item really. This will prevent the need to inject progress-tracking into all kinds of models when they could instead become an optional part of the foundational unit that all items share, _active effects_. 

I strongly believe this approach will be more flexible going forward, and also give us and users the chance to easily introduce temporary progress tracks through temporary effects that can be easily added onto existing actors at the click of a button. The potential of this approach will give both us and our users a great deal of creative freedom.

Furthermore, by having the progress data managed within the active effect model we will be able to manipulate it based on combat events. (For example, whenever a combat ends a progress track could be reset. This would work for the esper's brainwave clock).

## Changes

- Add support for progress tracks in active effects
- Add function to render an active effect to chat, FUActiveEffect.sendToChat
- Add enabled, id properties to the progress data model
- Add support for $pg macro to be used within effects.
- Refactor the progress track resolution on actor and item
- Add rules section for the active effect model
- Refactor the source info static ctor used in the inline actions
- Embed effect source information into InlineSourceInfo
- Add template for active effects rendered to chat messages
- Use labels in inline-damage, inline-resources for the damage pipeline

image

image

image

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -898,6 +898,7 @@
 		"ChatEvaluateAmountNoActor": "No reference to an actor provided",
 		"ChatEvaluateAmountNoSource": "No reference to a source provided",
 		"ChatEvaluateAmountNoItem": "No reference to an item provided",
+		"ChatEvaluateNoEffect": "No reference to an effect provided",
 		"ChatEvaluateNoSourceActor": "The given item must be added to an actor in order to evaluate",
 		"ChatEvaluateNoSkill": "The referenced skill was not found in the actor",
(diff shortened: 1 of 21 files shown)


spyrella:
Other use cases for this could possibly be:
- Predicate: effect only applied when clock is full.
- Duration: tracking amount of times an effect can apply before the active effect expires. (Ex: Divination has two uses before spell ends.)

Issue #301, opened 2025-04-19
Pilot Weapon Modules fields not using SchemaField data type


bblonski:
The following fields should be wrapped with the SchemaField object.

https://github.com/League-of-Fabulous-Developers/FoundryVTT-Fabula-Ultima/blob/69eeeb51a9d00123e48bcdb676274264f629bcca/module/documents/items/classFeature/pilot/weapon-module-data-model.mjs#L118-L124

Frozedon:
Fixed in #283

PR #306, opened 2025-04-20
Fix Pilot Weapon Modules fields not using SchemaField data type

bblonski:
Fix issues with Weapon Modules data schema causing them to not be treated like weapons when applying various bonuses. Issue #301


spyrella:
I'd wait for an opinion from @Shourn regarding this, he set up weapon module data model that specific way for a reason.

bblonski:
Any progress on this? I've fixed this locally but have to hold off on updates until it's fixed upstream.

PR #325, opened 2025-04-25
Fixed some compendium typos

Crosant:
Fixed Typos in Arcane Regeneration and Elemental Shard


spyrella:
Thanks for the fix, compendium sure is too big for one person to manage.

PR #357, opened 2025-05-09
Replaces absolute paths with relative paths

ravynhardcastle:
The CSS currently uses absolute paths for styling (fonts, images, etc.) This is fine in most cases. However, in the case where something is hosted in a location with a nested path segment (imagine someone with a url of `mywebsite.com/foundry/systems/projectfu/...`) this fails, as it attemps to pull from `mywebsite.com/systems/projectfu/...` path instead, which is invalid.

Anywhere that points to the absolute path `/systems/projectfu/styles` should be replaced with `..`. This PR is a simple find and replace to do just that. There is already precedence for this in the CSS, as the RPGAwesome import uses this format.

### DIFF
--- a/styles/css/projectfu.css
+++ b/styles/css/projectfu.css
@@ -2404,50 +2404,50 @@
 
 @font-face {
   font-family: "Yanone Kaffeesatz";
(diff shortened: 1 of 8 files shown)


spyrella (approved):

spyrella:
Thanks for that fix, merged.

Issue #396, opened 2025-05-16
Font in party sheet is not legible without pfu theme module


spyrella:
**Bug Description:**
Without the pfu theme module, font color appears illegible.

**Attachments (if applicable):**

![Image](https://github.com/user-attachments/assets/e0fe0789-25f9-47c5-8f00-cd42653e51c2)

Azurelol:
A recent submission addresses this. In the near future I hope a cascading style monarch descends from the heavens and unifies our warring style sheets.

PR #408, opened 2025-05-24
Fix/party sheet styling

Demaix:
fixes the issue #396

this is the current result: 
image


Demaix:
I messed up with the branches, sorry Q.Q Will close this Pull Request and fix it

PR #410, opened 2025-05-24
Added missing variable

Demaix:
The issue was/is that the party sheet referrences variables defined by the `Fabula Theme` module and not by the main repo. For a simple fix I put the variable for the text-color from the `Fabula Theme` module here. 

This is the end result.
image


Azurelol:
This looks fine to me a solution until we re-do the CSS classes in the nebulous feature.

Azurelol (approved):

PR #411, opened 2025-05-24
Fix verse dialog update procedure

### LINKED ISSUE #406: Verse automation description is lagging 1 option behind
**Bug Description:**
When using the verse automation and selecting a key or tone, the description in the preview dosn't update till the next change occurs and it shows the previous state. When press the start singing button, the correct verse will be posted in the chat
e.g.
first verse: key of fire + calm tone = hp recovery preview text
second verse: key of frost + calm tone = hp recovery preview text [text for first verse]
third verse: key of fire + calm tone = mp recovery preview text  [text for second verse]

**Steps to Reproduce:**
1. Create a chanter with at least 2 tones or key
2. Click on verse automation macro and set default values
3. change either key or tone
4. Observe the error occurring.

**Expected behavior:**
I expect the description to update to the currently selected key and tone.

In this Picture I have key of frost selected but it shows fire dmg type (should show frost dmg type instead)
Image

Demaix:
Changed the object update to updateDescription. Made key and tone optional parameters for when the value has changed. fixes #406

### DIFF
--- a/module/documents/items/classFeature/chanter/verses-application.mjs
+++ b/module/documents/items/classFeature/chanter/verses-application.mjs
@@ -105,8 +105,8 @@ export class VersesApplication extends FormApplication {
 		verse.app = this;
 
 		// Set predefined key and tone if provided in options
-		this.#defaultKey = options.predefinedKey ? verse.actor.items.get(options.predefinedKey) : this.keys[0];
-		this.#defaultTone = options.predefinedTone ? verse.actor.items.get(options.predefinedTone) : this.tones[0];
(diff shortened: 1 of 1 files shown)


Azurelol:
Sadly the issue persists on a fresh load, as seen here. It will not assign the key/tones correctly the first time:
image

If we solve that as well, then it will be on a good spot.

Demaix:
Updated PR title to your suggestion and I'm currently sitting on the problem. Forgot about that case

Azurelol (approved):
Excellent work.

PR #425, opened 2025-05-28
Fix/magic never hitting

### LINKED ISSUE #424: Spell Rolls always roll failure even if they would succeed
**Bug Description:**
When targeting an enemy, the spell shows an failure eventhough it would have succeeded.

**Steps to Reproduce:**
1. Add spell to character sheet.
2. Target enemy
3. Cast spell and beat it's mdef
4. Observe the error occurring.

**Expected behavior:**
It should show a success

Image

Demaix:
fixes #424
the type for the accuracy check only checked for accuracy but not magic, and spells have the type of magic. 
I tested it on both Fulgur and Flare and it worked so far.

### DIFF
--- a/module/checks/accuracy-check.mjs
+++ b/module/checks/accuracy-check.mjs
@@ -66,7 +66,7 @@ const onPrepareCheck = (check, actor, item, registerCallback) => {
  */
 const onProcessCheck = (check, actor, item) => {
 	const { type, critical, fumble } = check;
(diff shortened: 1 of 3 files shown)


Azurelol:
I took a good look at the pipelines and managed to fix it. Thank you for bringing it to our attention.

Azurelol:
If you would pull `dev` again and verify it's been fixed on your end as well?

PR #431, opened 2025-05-31
Update the system controls setup for v13

Azurelol:
The APIs to defining the system controls have breaking changes which needed to be addressed.


Azurelol:
There's still an error present when switching away from the PFU scene controls.
image

Azurelol:
Let's merge this and think about how to fix the current error next.

PR #435, opened 2025-06-02
fix: resolve image icon alignment issues within inline command buttons

Element-Re:
Should resolve the image wrapping and alignment issues. See below.
image


Element-Re:
Actually there's a remaining issue with the vertical alignment of the buttons themselves. Let me look into it further.

Azurelol (approved):
Excellent!

PR #436, opened 2025-06-03
Fix up the text editor enrichers

Azurelol:
- Migrate missing text editor enrichers to the newer hook
- Optimize said enrichers by using the new `onRender` property
- Consolidate the registration
- Deprecate inline affinity, which was replaced by inline type.


Azurelol:
Note that the inline drag-n-drop is currently borked; will wait until the actor sheet is on v2 to take another look at

PR #445, opened 2025-06-16
Translation fields adding

spechurkin:
Added fix for localization files in actor-party-section-settings.hbs in "FU.Settings" field. Also Added "Set Active" field in every localization file with DeepL translation to most languages.
Added hbs with localization on AwardResource in actor-party-sheet.mjs.


spechurkin:
Sure

Issue #378, opened 2025-05-12
Combat HUD Suggestions


spyrella:
**User Story:**
Suggestions for enhancing the combat hud.

**Suggestions:**
- [ ] 1. Hover over overflow effect menu (a feature from modern hud) for all hud types/pcs and npcs, for when users don't want to wait for effect to show up in marquee
- [ ] 2. Better compact mode for modern/mother hud
- [ ] 3. Add missing templates (fp/zenit)

**Attachments:**
1. Overflow effect menu
![Image](https://github.com/user-attachments/assets/c5bd78c3-12b4-44ad-8a93-8c84eb028641)

PR #446, opened 2025-06-17
Migrate CombatHUD To AppV2

Unarekin:
- Migrated CombatHUD to ApplicationV2
- Added missing FP and zenit resource partial templates.  These could probably be better but hey, now they exist and don't just break things.  Partially implements #378 
- Added 'clamp' HBS helper, and applied to clamp resource values from 0-999 in Mother template to avoid breaking the layout, which specifically expects 3 digit numbers.  And it's preeeeeeeeeetty unlikely a user will end up with a value greater than 999 being displayed on the template in the first place, since NPCs aren't shown on that one
- Left clicking on a status effect icon will:
  - Delete it if it is a temporary one (which is currently all of them actually displayed on the HUD) or
  - Disable it if it is not-temporary
- Right-clicking on a status effect icon will open a context menu with the option to disable, delete, or edit the effect.
- The above two options are only available if the current user is able to update or delete the effect, as appropriate to the action
- Added `pointer-events: none` to main HUD element for Mother and Modern templates, to allow mouse events to pass through to underlying elements (like the canvas)
- Explicitly added `pointer-events: initial` to `.window-button` and `.combat-row` elements, to ensure that these items handle pointer events even when the parent element has them disabled.
- Updated HUD related SceneControls to v13 style

### DIFF
--- a/module/projectfu.mjs
+++ b/module/projectfu.mjs
@@ -480,6 +480,10 @@ Handlebars.registerHelper('includes', function (array, value) {
 
 Handlebars.registerHelper('get', (map, key) => map?.[key]);
 
+Handlebars.registerHelper('clamp', (val, min, max) => {
+	return Math.max(Math.min(val, max), min);
+});
+
 /* -------------------------------------------- */
 /*  Ready Hook                                  */
 /* -------------------------------------------- */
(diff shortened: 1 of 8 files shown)


Azurelol (approved):
Nice work.

PR #453, opened 2025-06-23
Update InvocationSelectionApplication to v2

Unarekin:
- Update InvocationSelectionApplication to ApplicationV2
- Added styling to preserve original window background color from v1 because it did not look all that great with the default black
- Update reference to `TextEditor` to `foundry.applications.ux.TextEditor.implementation` as per deprecation

### DIFF
--- a/module/documents/items/classFeature/invoker/invocation-selection-application.mjs
+++ b/module/documents/items/classFeature/invoker/invocation-selection-application.mjs
@@ -1,23 +1,17 @@
 import { systemTemplatePath } from '../../../../helpers/system-utils.mjs';
+import FUApplication from '../../../../ui/application.mjs';
 import { WELLSPRINGS } from './invoker-integration.mjs';
 
-export class InvocationSelectionApplication extends foundry.applications.api.HandlebarsApplicationMixin(foundry.applications.api.ApplicationV2) {
+export class InvocationSelectionApplication extends FUApplication {
 	static DEFAULT_OPTIONS = {
-		form: {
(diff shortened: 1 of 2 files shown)


module/documents/items/classFeature/invoker/invocation-selection-application.mjs
```diff
@@ -1,20 +1,33 @@
+import { systemTemplatePath } from '../../../../helpers/system-utils.mjs';
 import { WELLSPRINGS } from './invoker-integration.mjs';
 
-export class InvocationSelectionApplication extends Application {
-	static get defaultOptions() {
-		return foundry.utils.mergeObject(super.defaultOptions, {
-			classes: ['form', 'projectfu', 'invocations-selection-app'],
-			width: 350,
-			height: 'auto',
+export class InvocationSelectionApplication extends foundry.applications.api.HandlebarsApplicationMixin(foundry.applications.api.ApplicationV2) {
```
Azurelol: Would you consider using `FUApplication`, which was recently defined for these applications? Though it may require some changes for it to apply more broadly.

Unarekin: I'll take a look at it

Unarekin: Now that I'm not napping or doing session 0: Went ahead and shifted over to FUApplication, cause future proofing sounds like a good idea.
And it's not like there's much actually on the class currently to cause problems.

PR #456, opened 2025-07-06
Consolidate item data model schemas

Azurelol:
- Define the base set of fields in the base schema
- Consolidate `source` field to be the same among all item types

### DIFF
--- a/module/documents/items/embedded-feature-data-model.mjs
+++ b/module/documents/items/embedded-feature-data-model.mjs
@@ -1,9 +1,9 @@
-import { FUBaseItemDataModel } from './item-data-model.mjs';
+import { FUItemDataModel } from './item-data-model.mjs';
 
 /**
  * @description Embeds a feature inside a data property
  */
-export class EmbeddedFeatureDataModel extends FUBaseItemDataModel {
+export class EmbeddedFeatureDataModel extends FUItemDataModel {
 	static defineSchema() {
 		const { StringField, SchemaField, BooleanField } = foundry.data.fields;
 		return Object.assign(super.defineSchema(), {
--- a/module/documents/items/embedded-feature-data-model.mjs
+++ b/module/documents/items/embedded-feature-data-model.mjs
@@ -1,15 +1,15 @@
+import { FUBaseItemDataModel } from './item-data-model.mjs';
+
 /**
  * @description Embeds a feature inside a data property
  */
-export class EmbeddedFeatureDataModel extends foundry.abstract.TypeDataModel {
+export class EmbeddedFeatureDataModel extends FUBaseItemDataModel {
 	static defineSchema() {
 		const { StringField, SchemaField, BooleanField } = foundry.data.fields;
-		return {
-			fuid: new StringField(),
+		return Object.assign(super.defineSchema(), {
 			summary: new SchemaField({ value: new StringField() }),
-			source: new StringField(),
 			isFavored: new SchemaField({ value: new BooleanField() }),
-		};
+		});
 	}
 
 	prepareDerivedData() {
--- a/module/documents/items/item-data-model.mjs
+++ b/module/documents/items/item-data-model.mjs
@@ -1,11 +1,35 @@
 /**
  * @description Base data model for items in the system
  */
-export class FUItemDataModel extends foundry.abstract.TypeDataModel {
+export class FUBaseItemDataModel extends foundry.abstract.TypeDataModel {
+	static defineSchema() {
+		const { StringField } = foundry.data.fields;
+		return {
+			fuid: new StringField(),
+		};
+	}
+
 	/**
 	 * @returns {FUPartialTemplate[]}
 	 */
 	get attributePartials() {
 		return [];
+	}
+}
+
+/**
+ * @description Standard data model for most items in the system
+ */
+export class FUItemDataModel extends FUBaseItemDataModel {
+	static defineSchema() {
+		const { SchemaField, StringField, HTMLField, BooleanField } = foundry.data.fields;
+		return Object.assign(super.defineSchema(), {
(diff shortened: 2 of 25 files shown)


module/documents/items/item-data-model.mjs
```diff
@@ -1,11 +1,35 @@
...
 }
+
+/**
+ * @description Standard data model for most items in the system
+ */
+export class FUItemDataModel extends FUBaseItemDataModel {
+	static defineSchema() {
+		const { SchemaField, StringField, HTMLField, BooleanField } = foundry.data.fields;
+		return Object.assign(super.defineSchema(), {
+			subtype: new SchemaField({ value: new StringField() }),
```
Shourn: Promoting subtype to apply to all items feels wrong.
As far as I can tell only consumables, heroics and treasure actually make use of it, and never to support mechanical resolution.

module/documents/items/embedded-feature-data-model.mjs
```diff
@@ -1,15 +1,15 @@
...
  * @description Embeds a feature inside a data property
  */
-export class EmbeddedFeatureDataModel extends foundry.abstract.TypeDataModel {
+export class EmbeddedFeatureDataModel extends FUBaseItemDataModel {
 	static defineSchema() {
 		const { StringField, SchemaField, BooleanField } = foundry.data.fields;
-		return {
-			fuid: new StringField(),
+		return Object.assign(super.defineSchema(), {
 			summary: new SchemaField({ value: new StringField() }),
```
Shourn, line 10: Why are you redefining `summary` and `isFavored` here?

module/documents/items/item.mjs
```diff
@@ -34,13 +34,22 @@ export class FUItem extends Item {
 		Hooks.callAll(FUHooks.DATA_PREPARED_ITEM, this);
 	}
 
+	/**
+	 * @override
+	 */
+	static migrateData(source) {
```
Shourn: This should not be in `FUItem`, but rather in `FUBaseItemDataModel`.
Any `DataModel` extending from `FUBaseItemDataModel` will also need to make sure to call `super.migrateData()`.

module/sheets/item-standard-sheet.mjs
```diff
@@ -53,6 +53,24 @@ export class FUStandardItemSheet extends FUItemSheet {
...
+	/**
+	 * @description Allow subclasses to dynamically configure render parts.
+	 * @param {HandlebarsRenderOptions} options
+	 * @returns {Record<string, HandlebarsTemplatePart>}
+	 * @protected
+	 */
+	_configureRenderParts(options) {
+		const parts = super._configureRenderParts(options);
+		switch (this.item.type) {
+			case 'effect':
```
Shourn: This looks like a bad solution to me.
If the `effect` sheet lacks some `parts`, why not just subclass the sheets for it and configure `PARTS` as required?

Shourn (changes requested):

module/documents/items/item-data-model.mjs
Azurelol: If that's the case, then we can do that. However as it has been for a long time, every "standard" item data model had those already. That's something I only noticed while going through the data models.

module/documents/items/embedded-feature-data-model.mjs
Azurelol: Because both the optional and class feature data models have it..

module/sheets/item-standard-sheet.mjs
Azurelol: It's an ugly solution, but I don't think it's particularly bad.

I don't disagree with you on that being cleaner and would be good to do eventually.

module/documents/items/item-data-model.mjs
Azurelol: I will go ahead and **only** add subtype to those data models.

module/documents/items/item.mjs
Azurelol: Er; I looked at some migration procedures (on the foundry v13 source), asked on the system dev channel, and believe this approach is a sound one. I am aware one can do it on a data model level but since all items' data models derive from this one, it should be a valid approach.

Azurelol: I am open to changing it if you insist; I merely thought this would be a cleaner approach.

Shourn: Yes, I insist.
You are building a class hierarchy for our data models.
That, in my opinion, is already a bad idea, but I won't stop you.
Putting logic that interacts specifically with how this new hierarchy stores its data into an unrelated class is a recipe for disaster and I will not stand for it.

module/sheets/item-standard-sheet.mjs
Shourn: Why not now? It should only be a couple of lines for the effect sheet since everything is inherited and some more for the sheet registration.

Shourn (changes requested):
GitHub didn't associate my other comments with the review.

module/sheets/item-standard-sheet.mjs
Azurelol: Very well.

module/documents/items/item.mjs
Azurelol: The ship has already sailed in how the data models on the system are structured; they are essentially already following a hierarchy by how their fields are defined. The intent here is to codify that in a clear manner, reduce duplication and prevent cases of fields being declared slightly differently on similar data models, or worse, being added to data models that don't need them. 

I disagree in stating that the item sheet class is unrelated to the item data model class, but I don't mind the migration being in either place so I will adjust it as you would prefer.

As an addendum, I am open to constructive feedback but I felt your tone here was a bit... hostile.

PR #465, opened 2025-07-18
Migrate Away from Socketlib Dependency

Unarekin:
Introduces `FUSocketHandler` class to abstract implementation of socket communication, removing explicit dependency on `socketlib` module.

## FUSocketHandler
- Instantiated in `game.projectfu` namespace
- Provides basic functionality to execute function handlers as a GM, other user(s), everyone, similar to how `socketlib` does
- Definitely not as feature-rich as `socketlib`, undoubtedly a little more error-prone for edge-cases
- Provides helper functions for existing socket calls:
  - requestStartTurn
  - requestEndTurn
  - requestTrade
  - requestZenitTransfer
  - studyEvent
  - showBanner
- Adds the following i18n keys because everyone likes meaningful error messages that actually indicate what went wrong:
  - `FU.RequestStartTurnNoActiveGM`
  - `FU.RequestEndTurnNoActiveGM`
  - `FU.RequestZenitTransferNoActiveGM`
  - `FU.RequestTradeNoActiveGM`
  - `FU.SocketHandlerAlreadyRegistered`

## Current Issues
The `requestTrade` functionality is not fully tested -- items can be transferred from a character actor to a party sheet, but a bug in the display of a party sheet is currently blocking testing of transferring items the other way around.  I'm confident it will work, since it's the same 

`showBanner` function is *functional* but needs its styling fix to get back to parity with the v12 implementation.  And maybe to actually use CSS variables so it can be themed via external means (Custom CSS, ProjectFU Theme)


lang/en.json
```diff
@@ -1462,6 +1462,10 @@
 			"ChangeWarning2": "Changing an Item's Fabula Ultima ID might break references to the Item! Proceed with caution and only if you are sure.",
 			"ChangeWarning3": "Do you want to proceed?",
 			"ChangeWarning1": "Proceed with caution!"
-		}
+		},
+		"RequestStartTurnNoActiveGM": "No GM online!  Unable to start turn.",
```
Shourn, line 1466: Have you added these keys to the translation google doc as well?

module/socket.mjs
```diff
@@ -1,68 +1,311 @@
...
+		if (!handler) return;
+
+		const message = Object.freeze({
+			...getBaseMessage(),
+			name,
+			users,
+			args,
+		});
+
+		Hooks.callAll(FUHooks.SOCKET_SEND_EVENT, message);
```
Shourn: Maybe use `Hooks.call()` here to allow for hooks to cancel the message dispatch?

module/socket.mjs
```diff
@@ -1,68 +1,311 @@
...
-	SOCKET.register(MESSAGES.ShowBanner, showBanner);
-	SOCKET.register(MESSAGES.RequestStartTurn, requestStartTurn);
-	SOCKET.register(MESSAGES.RequestEndTurn, requestEndTurn);
-	SOCKET.register(MESSAGES.RequestTrade, InventoryPipeline.requestTrade);
-	SOCKET.register(MESSAGES.RequestZenitTransfer, InventoryPipeline.requestZenitTransfer);
-	SOCKET.register(MESSAGES.StudyEvent, StudyRollHandler.onStudyEvent);
+		const message = Object.freeze({
+			...getBaseMessage(),
+			name,
+			users: game.users.reduce((prev, curr) => {
```
Shourn: Wouldn't `game.users.filter(user => user.active).map(user => user.id)` be more elegant than a reduce based solution?

module/socket.mjs
```diff
@@ -1,68 +1,311 @@
...
+	}
+
+	/**
+	 * Registers a message handler
+	 * @param {*} message
+	 * @param {*} handler
+	 */
+	register(message, handler) {
+		if (!(handler instanceof Function)) return;
+		if (!this.messageHandlers.has(message)) this.messageHandlers.set(message, handler);
```
Shourn: Would it be a good idea to throw or at least show an error if a message gets registered multiple times?

module/socket.mjs
```diff
@@ -1,68 +1,311 @@
...
+		await game.socket.emit(this.identifier, message);
+	}
+
+	/**
+	 * Execute a handler function as the first available GM user
+	 * @param {string} name - Name of the handler function.  Must be registered with {@link register}
+	 * @param  {...any} args - Arguments to pass to handler
+	 * @returns
+	 */
+	async executeAsGM(name, ...args) {
```
Shourn, line 122: I'm pretty sure that the actual message sending logic can be factored out and does not need ot be repeated three times.
It appears that only the recipient users change each time.

module/socket.mjs
```diff
@@ -1,68 +1,311 @@
...
+		const handler = this.messageHandlers.get(name);
+		if (!handler) return;
+
+		if (game.user.isGM) {
+			handler.apply(undefined, args);
+		} else {
+			const gm = game.users.find((user) => user.isGM && user.active);
```
Shourn: You should likely be using `game.users.activeGM` here, it's made for exactly this purpose.

module/socket.mjs
```diff
@@ -1,68 +1,311 @@
...
+		this.register(MESSAGES.RequestZenitTransfer, InventoryPipeline.requestZenitTransfer);
+		this.register(MESSAGES.RequestTrade, InventoryPipeline.requestTrade);
+		this.register(MESSAGES.StudyEvent, StudyRollHandler.onStudyEvent);
+
+		game.socket.on(this.identifier, (message) => {
+			// If this message is intended for only specific recipients and that is not us, do not process any further.
+			if (Array.isArray(message.users) && !message.users.includes(game.user.id)) return;
+
+			console.log('Socket message received:', message);
+			Hooks.callAll(FUHooks.SOCKET_RECEIVE_EVENT, message);
```
Shourn: I'm a bit unsure here, should hooks be run before or after the handler has executed? Especially since the check if the handler actually exists only happens after the hooks are run.

module/socket.mjs
```diff
@@ -1,68 +1,311 @@
...
+	}
+
+	/**
+	 * Requests to begin the turn for a given combatant.
+	 * @param {string} combatId
+	 * @param {string} combatantId
+	 */
+	async requestStartTurn(combatId, combatantId) {
+		try {
+			const gm = game.users.find((user) => user.isGM && user.active);
```
Shourn: Can use `game.users.activeGM` as well

module/socket.mjs
```diff
@@ -1,68 +1,311 @@
...
+	}
+
+	/**
+	 * Requests to end the turn for a given combatant
+	 * @param {string} combatId
+	 * @param {string} combatantId
+	 */
+	async requestEndTurn(combatId, combatantId) {
+		try {
+			const gm = game.users.find((user) => user.isGM && user.active);
```
Shourn: Can use `game.users.activeGM` as well

module/socket.mjs
```diff
@@ -1,68 +1,311 @@
...
+
+	/**
+	 *
+	 * @param {string} sourceActorId
+	 * @param {string} targetActorId
+	 * @param {number} amount
+	 */
+	async requestZenitTransfer(sourceActorId, targetActorId, amount) {
+		try {
+			const gm = game.users.find((user) => user.isGM && user.active);
```
Shourn: Can use `game.users.activeGM` as well

module/socket.mjs
```diff
@@ -1,68 +1,311 @@
...
+	/**
+	 * @param {String} actorId
+	 * @param {String} itemId
+	 * @param {Boolean} sale
+	 * @param {String} targetId
+	 * @returns {Promise<boolean|undefined>}
+	 */
+	async requestTrade(actorId, itemId, sale, targetId = undefined, modifiers = {}) {
+		try {
+			const gm = game.users.find((user) => user.isGM && user.active);
```
Shourn: Can use `game.users.activeGM` as well

module/socket.mjs
```diff
@@ -1,68 +1,311 @@
...
+		} catch (err) {
+			ui.notifications.error(err.message, { localize: true });
+		}
+	}
+
+	/**
+	 * @param {{actorUuid : String, targetUuids: String[], checkResult: Number}} data
+	 */
+	async studyRoll(data) {
+		this.executeAsGM(MESSAGES.StudyEvent, data);
```
Shourn: Pretty sure this should either be awaited or returned

module/socket.mjs
```diff
@@ -1,68 +1,311 @@
...
 	if (combat) {
 		const combatant = combat.combatants.get(combatantId);
-		if (combatant) {
-			await combat.endTurn(combatant);
-		}
+		if (combatant) await combat.endTurn(combatant);
 	}
 }
+
+function getBaseMessage() {
```
Shourn: I feel like this information (and possibly the list of intended recipients) should be made available to the handler functions, as an additional parameter added to the end of the args array maybe?

Shourn (changes requested):

lang/en.json
Unarekin: I hadn't, no.  I've actually been meaning to ask what the procedure is on adding translation keys, tbh.

module/socket.mjs
Unarekin: When we call the hook does kind of depend -- in another comment you suggest possibly switching to `Hooks.call` vs `Hooks.callAll`, which I do think is a good idea.  I can see a case for doing the same here, which would mean we'd want to call the hook *before* any associated handler, to give it a chance to be canceled.

On whether or not the hook should be called if there isn't a registered handler -- in my mind, the socket message being sent/received is a separate sort of event from it being handled, and what I would expect if I were using such a hook is that it would be called every time a message is sent or received.

Unarekin: ... though looking at it, I'm checking to make sure there's a handler before even sending the message, so maybe I'm just full of poo.

module/socket.mjs
Unarekin: Which information, to be clear?
Are you suggesting passing the triggering message wholesale to the handler function?

module/socket.mjs
Unarekin: Thinking on it, I do think this is a good idea.  Much of what I'd anticipated the hooks being for is a notification that a thing is happening, largely for debugging.

But, I can certainly see a potential use-case for canceling or mutating the message before it's sent or passed to its handler.

module/socket.mjs
Shourn: Maybe let me explain how I was thinking about this:

For the sending workflow I totally agree with:
1. Check for registered handler
2. Call hooks, possibly cancel sending
3. Send message

The receiving workflow on the other hand, I'd have assumed to work in the opposite direction (kinda):
1. Check for registered handler (disregard anyone sending garbage over our channel)
2. Call handler (the sending workflow was not cancelled, so we should probably take action here)
3. Call hooks, but with `callAll` since its too late for cancellation

Regarding your point about Sending/Receiving and Handling being separate, I generally do agree but:
Having a hook that gets called *before* we handle the message and the message being 100% mutable sounds like a bad thing to me.
If we want to call the hooks before the handling they should imho be passed a "deep-frozen" copy of the message.
That way we can be sure that the correct handler gets called with the correct data.

module/socket.mjs
Shourn: That might work as well.
I was more thinking about what you called "base data" here: `id, timestamp, sender`
Especially `sender` feels like something that could find some use.

Unarekin: Honestly that particular function was a method to ensure messages had standard data without repeating it in a few places.  Shifting over to have all messages sent through a single function obviates the need for `getBaseMessage` in particular.

That said, I am not against passing additional data to the handler functions.  Like you said, `sender` in particular could get some use.

module/socket.mjs
Unarekin: You know, I can get behind this.

module/socket.mjs
Unarekin: Ah, yeah.  Updated the handler calling bits to pass the actual message to the handler function as the last argument, in case anyone needs that info.  Figure the last argument would be the least inconvenient for function signatures that *don't* need it.

PR #476, opened 2025-08-06
Fix invocation wellspring description editor

Azurelol:


module/helpers/object-utils.mjs
```diff
@@ -25,6 +25,27 @@ function mergeRecursive(target, source) {
 	return [target, changed];
 }
 
+/**
+ * @param obj The object to resolve the property from
+ * @param path The path to the property, in dot notation
+ * @returns {undefined|*} The value of the property
+ */
+function getPropertyValueByPath(obj, path) {
```
Shourn: Why this function when Foundry already provides `foundry.utils.getProperty`, which has some additional optimizations and safeguards built in?

templates/feature/gourmet/cookbook-sheet.hbs
```diff
@@ -66,14 +66,15 @@
...
-                            {{localize (lookup @root/additionalData.tastes taste1)}} + {{localize (lookup
-                            @root/additionalData.tastes
-                            taste2)}}
+                            {{localize (lookup @root/additionalData.tastes taste1)}} + {{localize (lookup @root/additionalData.tastes taste2)}}
                         </label>
                     </legend>
                     <div>
-                        {{editor (lookup (lookup @root/enrichedHtml.combinations @key) 'effect') target=(concat
-                        'system.data.combinations.' @key '.effect') engine='prosemirror' button=true}}
+						{{formInput (get @root.dataFields (concat 'combinations.fields.' @key ".fields.effect"))
```
Shourn: Honestly not a fan of using the `formInput` helper for non-dynamic fields. iirc the helper exists to facilitate easier rendering of the very dynamic settings applications.
For our templates it causes loss of expressiveness while also losing a lot of flexibility.
I'd personally prefer just using the `<prose-mirror>` custom element directly.

templates/feature/invoker/invocations-sheet.hbs
```diff
@@ -40,11 +41,14 @@
                                 </label>
                             </div>
                             <div class="invocation-description">
-                                {{editor (lookup (lookup (lookup @root.enrichedHtml @key) "basic") "description")
-                                         target=(concat "system.data." @key ".basic.description") engine='prosemirror'
-                                         button=true}}
+                                {{formInput (get @root.dataFields (concat @key ".fields.basic.fields.description"))
```
Shourn: Same here as in the cookbook template.
Using `<prose-mirror>` instead of the `formInput` helper would improve the readability of what this part of the template tries to achieve.

module/helpers/object-utils.mjs
Azurelol: Agreed; will refactor to use it instead.

templates/feature/gourmet/cookbook-sheet.hbs
Azurelol: I am not sure what missed flexibility will be needed here or with invocations, but I haven't formed an opinion on which approach is better, or tried the other approach (I don't see examples of a direct `<prose-mirror>` element usage, only references in the editor handlebar templates.

If you wish us to try using prose-mirror here and elsewhere I have no reservations. (I am prioritizing getting these to work first).

templates/feature/invoker/invocations-sheet.hbs
Azurelol: See previous comment reply.

PR #492, opened 2025-08-19
Disable Token Drag Ruler Measurement

Unarekin:
Adds a check in the `init` Hook to disable the token drag ruler class, since Fabula does not use movement and positioning RAW.

image

Also adds a world setting to re-enable the ruler in case someone cares to use it.
image

image

Adds the following translation keys:
- FU.EnableDragRulerSettingGridlessLabel
- FU.EnableDragRulerSettingGridlessHint
- FU.EnableDragRulerSettingGriddedLabel
- FU.EnableDragRulerSettinGriddedHint
- FU.ShowTokenDragRulerKey


Unarekin:
Refactored to allow for an override keybind to show the drag ruler even when it is normally disabled.

image

lang/cn.json
```diff
@@ -1480,6 +1480,11 @@
 		"SheetOptionsTitle": "Document Sheets",
 		"SheetOptions": "Configure Sheets",
 		"SheetOptionsHint": "Configure various options for the system's sheets.",
+		"EnableDragRulerSettingGridlessLabel": "Enable Token Drag Ruler",
```
Azurelol: If it's not too troublesome, I think naming these keys with the domain first, so `DragRulerSettings...` first, would be best.

module/keybindings.mjs
```diff
@@ -1,8 +1,10 @@
 import { FUPartySheet } from './sheets/actor-party-sheet.mjs';
 import { Flags } from './helpers/flags.mjs';
+import { FUTokenRuler } from './ui/token-ruler.mjs';
 
-const KEYBINDINGS = Object.freeze({
+export const KEYBINDINGS = Object.freeze({
```
Azurelol: If we are exporting this, I would follow the convention and name them `FUKeybindings`.

Azurelol (changes requested):

module/keybindings.mjs
Unarekin: I don't have any issues renaming the export, but I do want to confirm that this *is* the naming convention so we're not going back and forth updating it unnecessarily.

The existing enums I'm aware of in the system don't consistently follow this, which is what gives me pause (SETTINGS does not, Traits does not, FUHooks does)

lang/cn.json
Unarekin: Sure, it's a pretty trivial change, and adding some sort of order to the 1500 lines of translation keys sounds good

module/keybindings.mjs
Azurelol: Aye, given the nature of thing I thought it best to follow that convention. I will look into updating Traits soon enough to do it.  Since SETTINGS doesn't have a prefix then it's okay for this not to have it for the time being either, as its the most closely related.

Azurelol (approved):

PR #493, opened 2025-08-20
feat: Refactor item list rendering

Shourn:
- refactor how item inlays in the actor sheet lists get rendered
- also slightly refactor action handling to allow moving logic out of the sheet into the data models

### DIFF
--- a/module/documents/items/common/progress-data-model.mjs
+++ b/module/documents/items/common/progress-data-model.mjs
@@ -375,4 +375,101 @@ export class ProgressDataModel extends foundry.abstract.DataModel {
 		}
 		return change;
 	}
+
+	/**
+	 * @typedef ClockUpdateOptions
+	 * @property {DirectProgressUpdateOptions} [direct] if provided checks for direct updates
+	 * @property {IndirectProgressUpdateOptions} [indirect] if provided checks for indirect updates
+	 */
+
+	/**
+	 * @typedef IndirectProgressUpdateOptions
+	 * @description an indirect update changes the value relative to its current value (increase/decrease operations)
+	 * @property {string} [dataAttribute="data-progress-action"] the data attribute signifying an indirect operation
+	 * @property {string} [attributeValueIncrement="increase"] which value signifies an increment operation
+	 * @property {string} [attributeValueDecrement="decrease"] which value signifies a decrement operation
+	 * @property {number} [changeAmountOverride] if provided changes will always be this amount irrespective of clicked mouse button or step size
+	 */
+
+	/**
+	 * @typedef DirectProgressUpdateOptions
+	 * @description a direct update changes the value to a specific new value
+	 * @property {string} [dataAttribute="data-segment"] the data attribute signifying a direct value assignment operation
+	 */
+
+	/**
+	 * @param {PointerEvent} event
+	 * @param {HTMLElement} target
+	 * @param {ClockUpdateOptions} [options]
+	 * @return Object
+	 */
+	getProgressUpdate(event, target, options = {}) {
+		/**
+		 * @param {string} dataAttribute
+		 * @return string
+		 */
+		function asDataSetAttribute(dataAttribute) {
+			return dataAttribute
+				.split('-')
+				.filter((value, index) => index > 0)
+				.map((value, index) => {
+					if (index === 0) return value;
+					return value.substring(0, 1).toUpperCase() + value.substring(1);
+				})
+				.join('');
+		}
+
+		const { direct, indirect } = options;
+
+		if (direct) {
+			const { dataAttribute = 'data-segment' } = direct;
+
+			const segmentElement = target.closest(`[${dataAttribute}]`);
+			if (segmentElement) {
+				const segmentDataSetAttribute = asDataSetAttribute(dataAttribute);
+				let newValue = Number(segmentElement.dataset[segmentDataSetAttribute]);
+
+				if (isNaN(newValue)) {
+					newValue = this.current;
+				} else {
+					if ((event.button === 0 && newValue === this.current) || event.button === 2) {
+						newValue -= 1;
+					}
+				}
+				return {
+					current: Math.clamp(newValue, 0, this.max || Number.MAX_SAFE_INTEGER),
+				};
+			}
+		}
+
+		if (indirect) {
+			const { dataAttribute = 'data-progress-action', attributeValueIncrement = 'increase', attributeValueDecrement = 'decrease', changeAmountOverride } = indirect;
+
+			const buttonElement = target.closest(`[${dataAttribute}]`);
+			if (buttonElement) {
+				const buttonDataSetAttribute = asDataSetAttribute(dataAttribute);
+
+				let amount = 0;
+				const operation = buttonElement.dataset[buttonDataSetAttribute];
+				if (operation === attributeValueIncrement) amount = 1;
+				if (operation === attributeValueDecrement) amount = -1;
+
+				if (Number.isInteger(changeAmountOverride)) {
+					amount *= changeAmountOverride;
+				} else {
+					if (event.button === 2) {
+						amount *= this.step;
+					}
+				}
+
+				const newValue = this.current + amount;
+
+				return {
+					current: Math.clamp(newValue, 0, this.max || Number.MAX_SAFE_INTEGER),
+				};
+			}
+		}
+
+		return {};
+	}
 }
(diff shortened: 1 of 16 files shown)


module/helpers/tables/table.mjs
```diff
@@ -0,0 +1,123 @@
...
+ * @typedef ColumnConfig
+ * @template T
+ * @property {boolean} [hideHeader]
+ * @property {number} [headerSpan]
+ * @property {"start", "center", "end"} [headerAlignment]
+ * @property {string, (() => string | Promise<string>)} [renderHeader]
+ * @property {string, ((T) => string | Promise<string>)} renderCell
+ */
+
+export class FUTable {
```
Azurelol: At first glance upon seeing these table class names, I thought they were data models or documents, but in fact are rendering functors that render a handlebars template (to be consumed piecemeal within a larger template)

Did you consider other names to explicitly declare this? For example something like `FUTableRenderer`?

module/sheets/actor-standard-sheet.mjs
```diff
@@ -149,20 +146,52 @@ export class FUStandardActorSheet extends FUActorSheet {
...
 	/**
 	 * @override
 	 */
 	_onClickAction(event, target) {
+		if (this.#dispatchClickActionToItem(event, target)) {
+			event.stopPropagation();
+			event.preventDefault();
+			return;
+		}
+
```
Azurelol: Do we not want to call the base method here?

module/helpers/tables/table.mjs
```diff
@@ -0,0 +1,123 @@
...
+
+		return foundry.applications.handlebars.renderTemplate('systems/projectfu/templates/table/fu-table.hbs', { tableId: this.#tableId, items, cssClass, columns, descriptions, expandedItems: this.#expandedItems });
+	}
+
+	/**
+	 * @param {HTMLElement} element
+	 */
+	bind(element) {
+		element.addEventListener('click', this.#clickHandler);
+	}
```
Azurelol: Would it be suitable to do the binding in the constructor to simplify the initialization of the table? (

Shourn: That would mean delaying instance creation for the tables until after first render for `ApplicationV2::element` to be initialized.
This won't work because context creation is already done by that point, meaning `renderTable` would have been called on `undefined`.
We could handle this by skipping the tables on first render and re-rendering immediately afterwards, but that sounds like a huge hack.

Azurelol: Ah. In that case, this was a sound decision.

module/sheets/actor-standard-sheet.mjs
Shourn: We can but right now it wouldn't do anything because `ApplicationV2::_onClickAction` is an intentionally empty extension point method.
If we were to introduce common logic in `FUActorSheet` for example we would need to call it.
That presents its own challenges though, since the base method returns void and if we wanted to honor that contract we'd need to either work out if an event has been handled by super or not call the super method if we handled it.

module/helpers/tables/table.mjs
Shourn: I'm not attached to the naming at all and will change it.
I was thinking more from a component standpoint, as is common in other web technologies, where a `Table` object usually represents a table in the HTML.

module/documents/items/common/progress-data-model.mjs
```diff
@@ -66,6 +66,15 @@ export class ProgressDataModel extends foundry.abstract.DataModel {
...
 	}
 
+	get clockSegments() {
+		return Array(this.max)
+			.fill(null)
+			.map((_, i) => ({
+				segment: i + 1,
+				filled: this.current > i,
+			}));
+	}
```
Azurelol, line 76: I believe  you can use the property `segments` here instead. (Maybe merge the approaches if  you prefer this one).

module/documents/items/misc/misc-ability-data-model.mjs
```diff
@@ -281,4 +280,44 @@ export class MiscAbilityDataModel extends FUSubTypedItemDataModel {
 			ItemPartialTemplates.progressClock,
 		];
 	}
+
+	updateResource(event, target) {
```
Azurelol, line 289: Would be good to leave comments here and in the next function to explain what they are doing since they are not directly referenced (only by the new template, I think)

module/documents/items/misc/misc-ability-data-model.mjs
```diff
@@ -281,4 +280,44 @@ export class MiscAbilityDataModel extends FUSubTypedItemDataModel {
...
+
+		if (target.closest('[data-segment]')) {
+			let newValue = target.closest('[data-segment]').dataset.segment;
+			if (event.type === 'contextmenu') {
+				newValue = newValue - 1;
+			}
+
+			return this.parent.update({
+				'system.progress.current': Math.clamp(newValue, 0, this.progress.max),
+			});
```
Azurelol: I believe this procedure and the one in rule data model can be factored out.

module/documents/items/ritual/ritual-data-model.mjs
```diff
@@ -172,4 +171,31 @@ export class RitualDataModel extends FUStandardItemDataModel {
 	get attributePartials() {
 		return [ItemPartialTemplates.standard, ItemPartialTemplates.ritual, ItemPartialTemplates.progressField];
 	}
+
+	updateRitualProgress(event, target) {
+		if (target.closest('[data-progress-action]')) {
```
Azurelol: I believe this one can be factored out as well.

module/documents/items/item-migrations.mjs
```diff
@@ -0,0 +1,21 @@
+import { SYSTEM } from '../../helpers/config.mjs';
+import { Flags } from '../../helpers/flags.mjs';
+
+function migrateFavoriteStatusToFlag(source) {
```
Azurelol, line 4: Woot!

module/documents/items/item.mjs
```diff
@@ -85,21 +87,8 @@ export class FUItem extends Item {
...
-			case 'treasure':
-			case 'material':
-			case 'artifact':
-			case 'accessory':
-			case 'consumable': {
-				return true;
-			}
-		}
-		return false;
+		return stashableTypes.has(this.type);
```
Azurelol, line 91: Understandable.

module/pipelines/effects.mjs
```diff
@@ -79,14 +78,14 @@ function createTemporaryEffect(owner, effectType, name) {
  */
 export async function onManageActiveEffect(event, owner, action) {
 	event.preventDefault();
-	const anchor = HTMLUtils.findWithDataset(event.target);
-	const listItem = anchor.closest('li');
+	const actionElement = event.target.closest('[data-action]');
```
Azurelol, line 81: Excellent change.

module/sheets/actor-standard-sheet.mjs
```diff
@@ -451,92 +477,67 @@ export class FUStandardActorSheet extends FUActorSheet {
...
-			if (actorId === this.actor.id) {
-				return await super._onDrop(ev);
-			}
-		}
-
-		// Proceed if the item is being dragged from the compendium/sidebar or other actors
-		const itemData = await this._getItemDataFromDropData(data);
+	async _onFirstRender(context, options) {
+		await super._onFirstRender(context, options);
+		this.#favoritesTable.activateListeners(this);
```
Azurelol, line 483: I wonder whether this could be refactored to be done all in one call; such as keeping a list of all registered tables for the sheet.

module/documents/items/common/progress-data-model.mjs
Shourn: I'm of two minds regarding this:
On the one hand, yes, we already had `progressArray` or the newer `segments`.
On the other hand, anything that relies on `generateProgressArray` must abide by its logic: the "is this segment filled" property being called `checked` and the array being in reverse order.

In this case I decided I'd rather have a new property that is more intuitive in its order and naming for the new tables while also avoiding breaking any other parts of the system.

module/documents/items/misc/misc-ability-data-model.mjs
Shourn: I'll try, I just hope the result is actually an improvement

module/sheets/actor-standard-sheet.mjs
Shourn: I was thinking about configuring the tables in the sheet options and building infrastructure for automatic instantiation and setup, but in the interest of time and the general size of this change I decided against it.

module/documents/items/common/progress-data-model.mjs
Azurelol: Very well. I can look into it after.

module/sheets/actor-standard-sheet.mjs
Azurelol: Understandable.

PR #496, opened 2025-08-22
ActorWellspringManager attempts to rerender actor sheet unnecessarily

Unarekin:
`ActorWellspringManager` rerenders its actor's sheet when the available wellsprings or current scene changes.  It does does so indiscriminately, attempting to rerender the sheet for *every* actor, which generates a warning for any actors for which the current user does not have Observer permission or higher.

This PR simply adds a check to make sure the sheet is rendered before attempting to rerender it, both to avoid this warning and to avoid trying to rerender sheets that aren't even rendered in the first place.


Azurelol:
Thanks!

PR #500, opened 2025-08-23
Adds Pixel Theme Option to Combat HUD

c-loaf:
Also:
- Changed "top" and "bottom" options for the hud window buttons from set values to CSS classes to allow easier customization and added those classes to the default styling
- Increased the distance from the default combat hud position and the top of the screen from 20 to 48 so the pixel hud's movement


module/helpers/templates.mjs
```diff
@@ -216,6 +216,14 @@ export const preloadHandlebarsTemplates = async function () {
 		'systems/projectfu/templates/ui/partials/combat-bar-ip-mother.hbs',
 		'systems/projectfu/templates/ui/partials/combat-bar-exp-mother.hbs',
 		'systems/projectfu/templates/ui/partials/combat-bar-zeropower-mother.hbs',
+		'systems/projectfu/templates/ui/partials/combat-bar-hp-pixel.hbs',
```
Unarekin, line 219: Preloading the template partials here is actually superfluous since they're included in the `templates` property of the PART on the `CombatHUD` application, which is the better place to define these partials.

Azurelol: Aye, ApplicationV2 greatly simplified the usage of partials.

PR #519, opened 2025-09-03
Add rule elements for active effects

Azurelol:
# Rule Element System

The goal of this feature is to provide a flexible, scalable system of automation for many cases which our current systems can't accommodate, and to do it in a fashion that is both easy to grasp, scale up, and that starts from the smallest elements and can be combined together to create a pleasant automation experience for our users.

I will be using a shorthand, RE, for them here.

_Note that this implementation was directly inspired by PF2es rule elements._

### Rule Triggers

Every rule element has a rule trigger which determines when the automation provided by the element kicks in. These triggers themselves can be changed at anytime, and are configurable. 

The triggers receive events that are dispatched to the actor their effect belongs to, during a conflict (combat).

- Combat: When a combat event is received, the rule element is triggered. The events range from start of combat, start of character turn, end of character turn, end of round, etc.
- Damage:  When a character receives damage.
- Resource Update: When one the character's resources changes (hp/mp/ip).
- Status: When a character gains/loses a status effect.
- Attack
- Spell
- Status
- Toggle (When the AE the RE is on is disabled/enabled)
- Resolve Check
- Perform Check
- Resource Expend
- Calculate Damage

### Rule Predicates

- Species
- Weapon
- Bond
- Faction Relationship
- Targeting
- Effect

### Rule Actions

- Apply Damage: Applies damage to the character.
- Update Resource: Add/Remove to the actor's resources.
- Inflict Status on Target: On a hit for a weapon/spell/skill, apply an active effect on the target.
- Adjust progress track (clock/resource points
- Adjust traits
- Modify Expense
- Modify Damage
- Notify
- Update Track (clock progress)
- Update Token
- Play Sound Effect

### Sample Use Cases

- Regeneration: At the start of a turn can be provided by a RE that triggers at the start of the character's turn and applies a resource change.
- Damage over time: This could be provided  by an RE that triggers at the end of a character's turn and applies damage.

### Supported Skills

These are a list of skills whose automation is now supported:

- Chimerist: Consume
- Mutant: Biophagy
- Darkblade: Agony
- Entropist: Absorb MP
- Spiritist: Healing Power
- Weaponmaster Counterattack
- Loremaster: Focused, Flash of Insight
- Esper: Cognitive Focus, Hypercognition

#### Examples

image


module/helpers/async-hooks.mjs
```diff
@@ -0,0 +1,81 @@
+/**
+ * AsyncHooks — A simple async/await–aware hook system.
+ *
+ * Usage:
+ *   AsyncHooks.on("myHook", async (data) => { ... });
+ *   await AsyncHooks.call("myHook", { foo: "bar" });
+ */
+export class AsyncHooks {
```
Shourn, line 8: This is a horrible idea and places an undue burden on any future maintainer.
I understand that Hooks being synchronous is a little awkward at times, but building a secondary Hook system that _mirrors_ the original API and uses the the same/similar names but with different semantics is not the way.
The Checks API already contains multiple examples of how to manage async operations triggered by Hooks, please have another look.

module/checks/checks.mjs
```diff
@@ -515,12 +517,25 @@ reapplyClickListeners();
  */
 const performCheck = async (check, actor, item, initialConfigCallback = undefined, onPerform = undefined) => {
 	const preparedCheck = await prepareCheck(check, actor, item, initialConfigCallback);
+	CommonEvents.performCheck(check, actor, item);
 	const roll = await rollCheck(preparedCheck, actor, item);
 	const result = await processResult(preparedCheck, roll, actor, item);
+	switch (check.type) {
```
Shourn: This logic should not be part of this method.
The Checks API is supposed to be modular and this code should be part of a module that only cares about `accuracy` and `magic` checks.

Since this is async code that needs to run after everything else it would make sense to enhance the `processCheck` Hook with the callback registration functionality already present in the `prepareCheck` Hook.

module/checks/common-sections.mjs
```diff
@@ -205,49 +206,69 @@ const opportunity = (sections, opportunity, order) => {
  * @param {FUItem} item
  * @param {TargetData[]} targets
  * @param {Object} flags
- * @param {CheckResultV2} checkData
- * @param {DamageData} damageData
+ * @param {CheckInspector} inspector
  */
-const targeted = (sections, actor, item, targets, flags, checkData = undefined, damageData = undefined) => {
+const targeted = (sections, actor, item, targets, flags, inspector = undefined) => {
```
Shourn: This should be renamed, this is much more than just targets.

module/documents/sub/typed-collection-field.mjs
```diff
@@ -0,0 +1,59 @@
...
+import { ModelCollection } from './model-collection.mjs';
+import { SubDocumentDataModel } from './sub-document-data-model.mjs';
+
+const { TypedObjectField, EmbeddedDataField } = foundry.data.fields;
+
+/**
+ * @desc A collection that houses sub-documents, which can have different types.
+ * @inheritDoc ModelCollection
+ */
+export class TypedCollectionField extends TypedObjectField {
```
Shourn: Can you please explain, and add documentation, what makes this different from and preferable to `new TypedObjectField(new TypedSchemaField(model.TYPES))`?

module/documents/effects/active-effect-config.mjs
```diff
@@ -158,6 +260,103 @@ export class FUActiveEffectConfig extends foundry.applications.sheets.ActiveEffe
...
+		await TypedCollectionField.addModel(this.document.system.rules.elements, type, this.document);
+		console.debug(`Added rule element`);
+	}
+
+	/**
+	 * @param {PointerEvent} event   The originating click event
+	 * @param {HTMLElement} target   The capturing HTML element which defined a [data-action]
+	 * @returns {Promise<void>}
+	 */
+	static async #deleteRuleElement(event, target) {
```
Shourn, line 284: This action should probably have a confirmation dialog.

module/documents/effects/active-effect-config.mjs
```diff
@@ -158,6 +260,103 @@ export class FUActiveEffectConfig extends foundry.applications.sheets.ActiveEffe
...
+			await ruleElement.delete();
+		}
+	}
+
+	/**
+	 * @param {PointerEvent} event   The originating click event
+	 * @param {HTMLElement} target   The capturing HTML element which defined a [data-action]
+	 * @returns {Promise<void>}
+	 */
+	static async #clearRuleElements(event, target) {
```
Shourn, line 302: This action appears to not do what its name suggests it would do, and if it did it would _definitely_ need a confirmation dialog.

module/documents/effects/active-effect-config.mjs
```diff
@@ -158,6 +260,103 @@ export class FUActiveEffectConfig extends foundry.applications.sheets.ActiveEffe
...
+		const re = this.document.system.getRuleElement(id);
+		await re.addRuleAction();
+	}
+
+	/**
+	 * @param {PointerEvent} event   The originating click event
+	 * @param {HTMLElement} target   The capturing HTML element which defined a [data-action]
+	 * @returns {Promise<void>}
+	 */
+	static async #removeRuleAction(event, target) {
```
Shourn, line 341: This action should probably have a confirmation dialog.

module/documents/effects/active-effect-config.mjs
```diff
@@ -158,6 +260,103 @@ export class FUActiveEffectConfig extends foundry.applications.sheets.ActiveEffe
...
+		const re = this.document.system.getRuleElement(id);
+		await re.addRulePredicate();
+	}
+
+	/**
+	 * @param {PointerEvent} event   The originating click event
+	 * @param {HTMLElement} target   The capturing HTML element which defined a [data-action]
+	 * @returns {Promise<void>}
+	 */
+	static async #removeRulePredicate(event, target) {
```
Shourn, line 369: This action should probably have a confirmation dialog.

module/documents/effects/active-effect-model.mjs
```diff
@@ -10,10 +12,23 @@ import { ProgressDataModel } from '../items/common/progress-data-model.mjs';
...
  * @property {ProgressDataModel} rules.progress It can be used for tracking a clock, a resource, a counter, etc.
  * @remarks The remaining property is initialized, and must be updated.
  */
 export class FUActiveEffectModel extends foundry.abstract.TypeDataModel {
+	/**
+	 * @type {SubDocumentMetadata}
+	 */
+	static get metadata() {
+		return {
+			...super.metadata,
```
Shourn: This looks weird, none of the superclasses have a static field `metadata`, likely an error?

module/documents/sub/sub-document-data-model.mjs
```diff
@@ -0,0 +1,205 @@
...
+	 */
+	get id() {
+		return this._id;
+	}
+
+	/**
+	 * The uuid of this document.
+	 * @type {string}
+	 */
+	get uuid() {
```
Shourn: I don't think calling this uuid is a great idea because `foundry.utils.fromUuid` and `foundry.util.fromUuidSync` will not work with this.

module/documents/effects/rule-element-data-model.mjs
```diff
@@ -0,0 +1,202 @@
...
+			}),
+			actions: new TypedCollectionField(RuleActionDataModel),
+			predicates: new TypedCollectionField(RulePredicateDataModel),
+			selector: new fields.StringField({ initial: 'initial', choices: Object.keys(FU.targetSelector) }),
+			enabled: new fields.BooleanField({ initial: true }),
+		});
+	}
+
+	static migrateData(source) {
+		if (source.trigger.selector) {
```
Shourn: This looks weird, why is it necessary to mutate the source of a nested data model?

module/documents/effects/rule-element-data-model.mjs
```diff
@@ -0,0 +1,202 @@
...
+		for (const action of this.actions) {
+			await action.renderContext(context);
+		}
+	}
+
+	/**
+	 * @returns {String}
+	 */
+	get header() {
+		return `${StringUtils.localize(this.trigger.schema.model.localization)}`;
```
Shourn: Wrapping the method call in a template string appears unnecessary.

module/documents/effects/rule-element-data-model.mjs
```diff
@@ -0,0 +1,202 @@
...
+	async prepareContext(context) {
+		for (const action of this.actions) {
+			await action.renderContext(context);
+		}
+	}
+
+	/**
+	 * @returns {String}
+	 */
+	get header() {
```
Shourn: Please document that this method is used for handlebars templates, it looks unused/deletable otherwise.

module/checks/common-events.mjs
```diff
@@ -403,38 +492,151 @@ function progress(document, progress, action, increment = undefined, source = un
...
+/**
+ * @typedef InitializeCheckEvent
+ * @property {CheckConfigurer} configuration
+ * @property {CharacterInfo} source
+ * @property {InlineSourceInfo} sourceInfo
+ * @property {Object} flags
+ * @remarks Emitted when a check is about to be rendered.
+ */
+
+async function initializeCheck(configuration, actor, item) {
```
Shourn, line 575: This event appears to be inconsistently applied.
For item types with initial configuration callbacks it is dispatched at the end of said callbacks, but for `display` "checks" it is dispatched before the callback.

Wouldn't it make more sense to move this into the `Checks.prepareCheck` method and dispatch it if a config callback was provided and after the callback has executed?
That way it would not be necessary to sprinkle event dispatches across the entire system.

module/helpers/character-info.mjs
```diff
@@ -0,0 +1,103 @@
...
+	/** @type {CharacterInfo} */
+	return {
+		actor: combatant.actor,
+		token: combatant.token,
+		combatant: combatant,
+		disposition: resolveDisposition(combatant.actor, combatant.token),
+	};
+}
+
+export const CharacterInfo = Object.freeze({
```
Shourn: Weird mix of top level functions and functions defined in the export object here.
Please choose one style and clean this up.

module/helpers/foundry-utils.mjs
```diff
@@ -173,4 +233,26 @@ export default class FoundryUtils {
...
+			value: key,
+		}));
+	}
+
+	/**
+	 * @param {String} text
+	 * @param {Object} context
+	 * @returns {Promise<*>}
+	 */
+	static async enrichText(text, context) {
```
Shourn, line 255: If this is supposed to be the future preferred way to enrich text please mark `text-editor.mjs` as deprecated for removal.

module/helpers/inline-helper.mjs
```diff
@@ -247,25 +253,11 @@ function appendVariableToAnchor(anchor, key, expression, localization = 'FU.Vari
...
 function toBase64(value) {
-	try {
-		const string = JSON.stringify(value);
-		const bytes = new TextEncoder().encode(string);
-		const binString = Array.from(bytes, (byte) => String.fromCodePoint(byte)).join('');
-		return btoa(binString);
-	} catch (e) {
-		return null;
-	}
+	return StringUtils.toBase64(value);
```
Shourn: Why move this to `StringUtils` but then keep the function?

module/helpers/inline-helper.mjs
```diff
@@ -247,25 +253,11 @@ function appendVariableToAnchor(anchor, key, expression, localization = 'FU.Vari
...
 function fromBase64(base64) {
-	try {
-		const binString = atob(base64);
-		const uint8Array = Uint8Array.from(binString, (m) => m.codePointAt(0));
-		const decodedValue = new TextDecoder().decode(uint8Array);
-		return JSON.parse(decodedValue);
-	} catch (e) {
-		return null;
-	}
+	return StringUtils.fromBase64(base64);
```
Shourn: Why move this to `StringUtils` but then keep the function?

module/helpers/object-utils.mjs
```diff
@@ -43,8 +43,13 @@ function cleanObject(obj) {
 	return Object.fromEntries(Object.entries(obj).filter(([_, v]) => v !== undefined));
 }
 
+function selectKeys(record, keys) {
```
Shourn: Some documentation would be helpful, especially here where the method does something that many JS devs already know by another name (`pick` in this case).

styles/static/icons/affinities/air.png
Shourn, line 1: Why did you copy all these?

templates/item/hoplosphere/hoplosphere-effect-choice-and-amount.hbs
```diff
@@ -3,7 +3,7 @@
 	<label class="resource-label-sm">
 		{{localize templateData.choice.label}}
 		<select name="{{fieldPrefix}}.{{templateData.choice.name}}" class="resource-inputs select-dropdown-full">
-			{{selectOptions templateData.choice.choices selected=(lookup effect templateData.choice.name) localize=true}}
+			{{selectOptions templateData.choice.qualifiedChoices selected=(lookup effect templateData.choice.name) localize=true}}
```
Shourn: Global rename error, needs revert.

templates/item/hoplosphere/hoplosphere-effect-single-choice.hbs
```diff
@@ -2,6 +2,6 @@
 <label class="resource-label-sm">
 	{{localize templateData.label}}
 	<select name="{{fieldPrefix}}.{{templateData.name}}" class="resource-inputs select-dropdown-full">
-		{{selectOptions templateData.choices selected=(lookup effect templateData.name) localize=true}}
+		{{selectOptions templateData.qualifiedChoices selected=(lookup effect templateData.name) localize=true}}
```
Shourn: Global rename error, needs revert.

module/documents/sub/sub-document-data-model.mjs
```diff
@@ -0,0 +1,205 @@
...
+				required: true,
+				blank: false,
+				initial: this.TYPE,
+				validate: (value) => value === this.TYPE,
+			}),
+		};
+	}
+
+	/** @inheritdoc */
+	renderContext() {
```
Shourn: `renderContext` is a misleading name for this method, it should probably be called either `prepareContext` or `prepareRenderContext`.

Shourn (changes requested):

styles/static/icons/affinities/air.png
Azurelol: These were copied to better align with the rest of the icons in `styles/static/icons/*`; it's something I planned to bring up on the channel regarding common icon locations.

Azurelol: I hadn't yet decided whether to go along with that and update the references to the previous location across or abandon the endeavor.

module/helpers/object-utils.mjs
Azurelol: Will do.

templates/item/hoplosphere/hoplosphere-effect-choice-and-amount.hbs
Azurelol: Ack. Will fix.

templates/item/hoplosphere/hoplosphere-effect-single-choice.hbs
Azurelol: Ack. Will fix.

module/documents/sub/sub-document-data-model.mjs
Azurelol: I struggled to come up with a name for it; I will go with one of your suggestions.

module/documents/effects/active-effect-config.mjs
Azurelol: Tis done.

module/documents/effects/active-effect-config.mjs
Azurelol: Damn, not sure where the actual clear procedure went. Fixed it and added a confirmation dialog.

module/documents/effects/active-effect-config.mjs
Azurelol: Tis done.

module/documents/effects/active-effect-config.mjs
Azurelol: Tis done.

module/documents/effects/active-effect-model.mjs
Azurelol: Will remove the super. Indeed an oversight.

module/documents/effects/rule-element-data-model.mjs
Azurelol: This one is not needed anymore, will remove; there's a few of these WIP migration methods around the RE data models as I prototyped their data structures.

module/documents/effects/rule-element-data-model.mjs
Azurelol: Indeed.

module/documents/effects/rule-element-data-model.mjs
Azurelol: Will do.

module/helpers/inline-helper.mjs
Azurelol: I had punted off removing these, but it's done now.

module/helpers/inline-helper.mjs
Azurelol: I had punted off removing these, but it's done now.

module/helpers/character-info.mjs
Azurelol: Fixed.

module/checks/common-events.mjs
Azurelol: Good points. I will try that.

module/checks/common-sections.mjs
Azurelol: Will do.

module/checks/checks.mjs
Azurelol: It's done.

module/helpers/async-hooks.mjs
Azurelol: I have gone over other issues in the checks API related to this usage and simplified it, and reused the same pattern employed when possible.

For the use cases used in the 4 events that make use of this API, I did not find a substitute that wouldn't complicate the checks API further. I don't want REs to directly subscribe or register to the checks pipeline directly, but instead just to the events and have that layer of abstraction, in the simplest API possible.

module/documents/sub/typed-collection-field.mjs
Azurelol: It's meant to be used exclusively for the SubDocumentDataModel. 

I have renamed the class to be unambiguous regarding that.

module/documents/sub/sub-document-data-model.mjs
Azurelol: Sure. It's only used internally anyways.

module/helpers/foundry-utils.mjs
Azurelol: At this time I have no such plan. I just felt it is cumbersome to directly invoke the text editor API in such a way.

PR #539, opened 2025-09-14
Custom Weapons

Shourn:


module/helpers/tables/weapons-table-renderer.mjs
```diff
@@ -4,27 +4,58 @@ import { CommonColumns } from './common-columns.mjs';
 import { systemTemplatePath } from '../system-utils.mjs';
 import { FU } from '../config.mjs';
 
+const weaponTypes = new Set(['weapon', 'customWeapon']);
```
Azurelol, line 7: Seems this could be defined statically elsewhere to re-use.

module/sheets/item-sheet.mjs
```diff
@@ -243,6 +243,21 @@ export class FUItemSheet extends api.HandlebarsApplicationMixin(sheets.ItemSheet
...
+	}
+
+	disableElement(element) {
+		element.classList.add('disabled');
+		if (element instanceof HTMLInputElement || element instanceof HTMLTextAreaElement) {
+			element.readOnly = true;
+		} else {
+			element.disabled = true;
+		}
+		element.parentElement.dataset.tooltip = game.i18n.localize('FU.DisabledByActiveEffect');
```
Azurelol, line 260: Perhaps move to the foundry utils class?

module/sheets/item-sheet.mjs
```diff
@@ -488,12 +503,12 @@ export class FUItemSheet extends api.HandlebarsApplicationMixin(sheets.ItemSheet
...
+		const data = super._prepareSubmitData(event, form, formData, updateData);
 		// Prevent submitting overridden values
-		const overrides = foundry.utils.flattenObject(this.item.overrides);
-		for (let k of Object.keys(overrides)) {
-			delete data[k];
+
+		for (let k of Object.keys(foundry.utils.flattenObject(this.item.overrides))) {
+			foundry.utils.deleteProperty(data, k);
 		}
 		return data;
```
Azurelol, line 513: What is this change about? (Not that I am against it or anything)

system.json
```diff
@@ -294,85 +294,88 @@
...
-				"htmlFields": ["biography.value", "biography.gm"]
+				"htmlFields": ["description"]
 			},
 			"party": {
-				"htmlFields": ["biography.value", "biography.gm"]
+				"htmlFields": ["description"]
 			},
 			"stash": {
-				"htmlFields": ["biography.value", "biography.gm"]
+				"htmlFields": ["description"]
```
Azurelol, line 306: Whoops. When I migrated to the newer config style I forgot all along to change this from the template.

module/sheets/item-sheet.mjs
Shourn: This basically fixes something that was broken by the AppV2 migration.
What it does and the intent has not changed, but the name of the base method changed from AppV1 to AppV2.
This functionality basically replicates what the ActorSheet does by default, namely cleaning fields that get overridden by active effects from the submit data.
Without this the modified value would get written to the database leading to a compounding problem as we have seen recently with the rest function.
Since Foundry does not assume that ActiveEffects get applied to items the ItemSheet baseclass lacks this functionality.

module/sheets/item-sheet.mjs
Shourn: I like it to be here in the sheet.
I think it aids discoverability if the functionality is near where it is supposed to be used.

module/helpers/tables/weapons-table-renderer.mjs
Shourn: That would make this class more brittle and harder to change.
I know this particular constant exists twice right now, but that is because the context is very different.
Here it defines which item types should be shown in the table, in `choose-weapon-dialog.mjs` it defines what the dialog should consider a "weapon".
There's a difference in semantics there and forcing a unified constant might cause one or the other to break if a new value were to be added to the set without prior thought.

module/sheets/item-sheet.mjs
Azurelol: I do not think discoverability is an issue here, but I won't press since this behavior is not needed elsewhere.

module/helpers/tables/weapons-table-renderer.mjs
Azurelol: I do not think there would be more weapon types anywhere in the future, right? Though it's a small matter.

Shourn: It's an unlikely scenario, but if we ever decide to "ditch" the idea of "class features" and elevate them to proper item types it's highly likely that the dialog will consider "weapon modules" as weapons while it might make sense to group all module types including weapon modules into a table.

PR #550, opened 2025-10-06
Feature/technospheres

Shourn:
# Introduction

This PR introduces new infrastructure for `PseudoDocuments`, as well as support for the Technosphere rules from Atlas Techno Fantasy.  
The Technosphere implementation serves as a benchmark for the fitness for purpose of the new infrastructure.

# New Infrastructure

## PseudoDocuments

This PR introduces a range of new infrastructure for managing `PseudoDocuments` embedded in primary `Documents`, as well as `PseudoDocuments` embedded within other `PseudoDocuments`.  
The most important parts of this infrastructure are found in `/module/documents/pseudo`:

- `PseudoDocument` in `pseudo-document.mjs`
- `PseudoDocumentEnabledTypeDataModel` in `pseudo-document-enabled-type-data-model.mjs`
- `EnablePseudoDocumentsMixin` in `enable-pseudo-documents-mixin.mjs`
- `PseudoDocumentCollectionField` in `pseudo-document-collection-field.mjs`
- `PseudoDocumentCollection` in `pseudo-document-collection.mjs`
- `PseudoDocumentTypeDataField` in `pseudo-document-type-data-field.mjs`

### PseudoDocument

A `PseudoDocument` is a specialized implementation of `foundry.abstract.DataModel`, made to mimic the behaviour of `foundry.abstract.Document`.

This new model is needed because Foundry has very strong hardcoded assumptions of what a `Document` is.
The most relevant of these is that a `Document` is either stored in one of the primary collections (`Actors`, `Items`, etc.) or embedded into another `Document`.
Which `Documents` may be embedded into which other `Documents` is defined at the `BaseDocument` level (`BaseActor`, `BaseItem`, etc.).
Because `BaseDocuments` are a server-side construct, in contrast to `ClientDocumentMixin`-subclasses like `Actor` and `Item`, system or module code cannot change their data or behaviour.

In many aspects `PseudoDocument` is a combination of `Document` and `ClientDocumentMixin`, which is possible because it is an entirely client-side construct.
`PseudoDocument` is designed as a base class for subclasses like `PseudoItem` or `PseudoActiveEffect` that mimic a certain type of client document, but can conceivably also be used for types with no connection to existing Foundry document types.
Many methods are either direct copies or only slightly adjusted versions of those found on either `foundry.abstract.Document` or `ClientDocument`/`ClientDocumentMixin`.
This is done to present an API as close as possible to those of normal documents.
As result conversion between a `Document` and its corresponding `PseudoDocument` is easy and sheets work for both `Document` and `PseudoDocument` alike.

Unlike `Documents`, `PseudoDocuments` may be embedded into `TypeDataModels`, thereby allowing arbitrary nesting.
To give a practical example, a custom weapon (`Item`) can have an embedded mnemosphere (`PseudoItem`), which has embedded skills (`PseudoItem`) that have embedded effects (`PseudoActiveEffect`).

Of particular importance is the `PseudoDocument#_gatherChangeData` method, which is responsible for figuring out the "root" change data which has to submitted to the backend.
The "root" change data consists of the complete content of the first array found, with the actual change data being contained in the relevant "branches".
This procedure is necessary because Foundry is at present not able to do proper differential updates on arrays.

### PseudoDocumentEnabledTypeDataModel

This subclass of `foundry.abstract.TypeDataModel` provides API mirroring `foundry.abstract.Document` for use by `EnablePseudoDocumentMixin`.
It is also responsible for initializing any `PseudoDocumentCollectionFields` / `PseudoDocumentCollections` that are defined by the data model.

Document types that want to support nested `PseudoDocuments` are required to use this as the base class for their `DataModel`.

### EnablePseudoDocumentMixin

This mixin mostly patches `Document` and `ClientDocument` methods to enable natural usage of existing API in the context of `PseudoDocuments`.

### PseudoDocumentCollectionField & PseudoDocumentCollection

These two classes are closely related and allow creating embedded collections both as part of `PseudoDocument` subclasses or as part of a `PseudoDocumentEnabledTypeDataModel`.
While `PseudoDocumentCollectionField` is a subclass of `foundry.data.field.ArrayField`, the initialized value is an instance of `PseudoDocumentCollection`.

### Design Notes

My primary design goal with `PseudoDocuments` was being able to embed any type of item into any other kind of item, while keeping surrounding infrastructure intact.
I found it especially important that item sheets would _just work_, no matter if it was a "real" item or a `PseudoDocument` item.

# PseudoDocument Implementations

Specializations of `PseudoDocument`, like `BasePseudoItem` and `BasePseudoActiveEffect`, **have** to define three things:

- `static documentName = "<documentName>";`
  - when implementing a `PseudoDocument` for a core Foundry document the name has to be the same.
- `static defineSchema(){}`
  - when implementing a `PseudoDocument` for a core Foundry document it is best to copy the schema definition of the base document and then selectively remove fields that have no relevance to the pseudo document.
- `static metadata = {};`
  - the object has three required fields, which can be inherited from `PseudoDocument` or overridden as needed:
    - `label`: This should always be overridden and should be a translation key for the document type.
    - `hasTypeData`: This defaults to `false` and has to be overridden to `true` for documents with subtypes, like items.
    - `coreTypes`: should contain the array of core Foundry subtypes if they exist otherwise `[]`.

Behaviour of the base document should be implemented on the respective `BasePseudoDocument`, see `BasePseudoActiveEffect` for an example.

## Behaviour Mixins

Since `PseudoDocuments` and `ClientDocuments` cannot share the same class hierarchy I had to find another solution to share behaviour.
The solution I found was inspired by how Foundry itself shares behaviour across unrelated class hierarchies, namely Mixins.

Each `ClientDocument`/`PseudoDocument` pair has an associated `BehaviourMixin` that implements system behaviour.
If absolutely necessary, this behaviour can be overridden but for maintainability purposes each document pair should behave the same.

# New Item Types

This PR introduces three new item types: `hoplosphere`, `mnemosphere` and `mnemosphereReceptacle`.

## Hoplospheres

The Hoplosphere data model itself is relatively straightforward, but it contains an array of `ChangeDataModels` that will require some explanation.

First of all, Active Effects are not a good choice for modeling the changes a hoplosphere may contain.
In a worst case scenario you'd need 3 independent Active Effects for a single coagulation level:

- one that should apply to the armor the hoplosphere is socketed in.
- one that should apply to the custom weapon the hoplosphere is socketed in.
- one that applies to the actor, regardless of the item type the hoplosphere is socketed in.

This kind of complexity is prohibitive to laymen users who just want to homebrew a single hoplosphere.

For this reason I opted for a similar approach to what we already have with the guided effect creator.
The predefined effects contain all the required information, but are defined as plain objects.
These plain objects are then used to create `DataModel` implementations at runtime, which are then used in the `TypedSchemaField` of the effects array.

Importantly, the change application functions of the `ChangeDataModels` are responsible for adding any changed keys to the `overrides` object on either actor or item.
This is necessary to prevent bonuses stacking with themselves after changes to the affected data models.

## Mnemospheres

Mnemespheres themselves are little more of a collection of other items, most likely skills.
They provide some utility and contain logic for when a nested item should show up on the actor sheet.

## Mnemosphere Receptacle

Mnemosphere Receptacles are simply containers for slotted Mnemospheres.
Their primary usecase is the "Integrated Mnemospheres" optional rule described on p138 of the Techno Fantasy Atlas.

# Changed Item Types

Both `ArmorDataModel` and `CustomWeaponDataModel` were changed to extend from `PseudoDocumentEnabledTypeDataModel`.
They now contain an `items` collection and fields to define the type of slots present on the item.

# Technospheres User Experience

Technospheres work by dragging and dropping them from any item list onto an open slot in the item sheet.

Because of the fiddly nature of items in items, editing technosphere sheets that belong to an actor or item is disabled by default.
It can be re-enabled through the three-dot-menu in the sheet header.

Slotted Mnemospheres only inherit skills with skill level 1 or greater to the actor to not clutter the sheet.


module/documents/actors/actor.mjs
```diff
@@ -244,15 +243,11 @@ export class FUActor extends Actor {
 	 * @override
 	 */
 	*allApplicableEffects() {
-		for (const effect of super.allApplicableEffects()) {
```
Azurelol, line 247: How come the change from the allApplicableEffects?

module/documents/items/hoplosphere/hoplosphere-data-model.mjs
```diff
@@ -0,0 +1,582 @@
...
+ * @property {(item: FUItem, actor: FUActor) => void} apply "this" will be bound to a DataModel defined by "schema"
+ * @property {() => DataSchema} defineSchema
+ * @property {boolean} [weaponOnly] for chat display purposes
+ * @property {(data: object) => void} prepareLocalization preprocess data for localization
+ */
+/**
+ *
+ * @type {Record<string, HoplosphereChangeDefinition>}
+ */
+const HOPLOSPHERE_CHANGE_TYPES = {
```
Azurelol, line 47: I would like to see these moved to a different file.

module/documents/actors/actor.mjs
Shourn: With items inside items we have to recurse through all items to gather all applicable effects.
It made more sense to me for this logic to be implemented in the items, especially since different item types can decide when all or some of their own effects are transferred.

module/documents/items/hoplosphere/hoplosphere-data-model.mjs
Shourn: To me this is an implementation detail of hoplospheres.
Moving them to another file would require to export them, making them accessible everywhere.
I don't see the value here.

module/documents/items/item.mjs
```diff
@@ -1,12 +1,16 @@
...
+import { ItemBehaviourMixin } from './item-behaviour-mixin.mjs';
 
-const stashableTypes = new Set(['accessory', 'armor', 'consumable', 'shield', 'treasure', 'weapon']);
+/**
+ * @typedef KeyboardModifiers
+ * @property {boolean} shift
+ * @property {boolean} alt
+ * @property {boolean} ctrl
+ * @property {boolean} meta
+ */
```
Azurelol: This typedef can be found on another file (though I should have moved it to `typedefs.mjs`.

module/documents/pseudo/pseudo-document-enabled-type-data-model.mjs
```diff
@@ -0,0 +1,137 @@
+import { PseudoDocumentCollectionField } from './pseudo-document-collection-field.mjs';
+
+export class PseudoDocumentEnabledTypeDataModel extends foundry.abstract.TypeDataModel {
```
Azurelol, line 11: Would like to see a comment for the usage of the class here.

Azurelol, line 11: And I assume this one can't use the enable pseudo document mixin?

module/documents/pseudo/pseudo-document.mjs
```diff
@@ -0,0 +1,751 @@
...
+	 * Return a reference to the parent Collection instance that contains this Document.
+	 * @this {PseudoDocument}
+	 * @returns {Collection|null}
+	 */
+	get collection() {
+		return this.parent[this.parentCollection];
+	}
+
+	/**
+	 * The Document this PseudoDocument belongs to. May be nested multiple layers deep.
```
Azurelol, line 167: Copy error in comment.

module/documents/pseudo/pseudo-document.mjs
```diff
@@ -0,0 +1,751 @@
...
+	 * @return {null|foundry.abstract.Document}
+	 */
+	get parentFoundryDocument() {
```
Azurelol, line 170: Wondering whether it would be more intuitive to name this one `parentDocument` and the other `parentPseudoDocument`.

module/pipelines/effects.mjs
```diff
@@ -99,13 +99,23 @@ export async function onManageActiveEffect(event, owner, action) {
...
+		case 'edit': {
+			const effect = resolveEffect();
+			return effect.sheet.render(true, { editable: owner === effect.parent });
+		}
 		case 'delete': {
 			const _effect = resolveEffect();
 			if (canBeRemoved(_effect)) {
-				sendToChatEffectRemoved(_effect, owner);
-				return _effect.delete();
+				if (
```
Azurelol, line 109: Hmm. I am on the fence on whether we want a confirmation for deleting active effects when the action to get there is already quite deliberate. I guess we can see how it goes and if it becomes annoying for users we can walk it back.

module/projectfu.mjs
```diff
@@ -329,6 +359,17 @@ Hooks.once('setup', () => {});
...
+Handlebars.registerHelper('times', function (n, block) {
+	let accum = '';
+	for (let i = 0; i < n; ++i) {
+		block.data.index = i;
+		block.data.first = i === 0;
+		block.data.last = i === n - 1;
+		accum += block.fn(this);
+	}
+	return accum;
+});
```
Azurelol: Let's move this to the `FUHandlebars` file alongside the other helpers.

module/documents/actors/actor.mjs
Azurelol: Understandable.

module/documents/items/hoplosphere/hoplosphere-data-model.mjs
Azurelol: When I open the file for the `HoplosphereDataModel`, it is only until the very end of the file, 400+ lines in that I see the actual model. 

While I do think it would be cleaner to see those on their own file, I will not press on this.

templates/item/partials/item-armor-description.hbs
```diff
@@ -0,0 +1,16 @@
+<div class="tab description {{tab.cssClass}}" data-group="primary" data-tab="description">
+	<fieldset class="desc title-fieldset resource-content">
+		{{> "projectfu.technospheres.slots" slots=slots }}
```
Azurelol, line 3: I believe this should go in the attributes section. Though I can move it there myself at a later time.

module/documents/pseudo/pseudo-document.mjs
Shourn: I don't understand, please explain?

module/documents/pseudo/pseudo-document.mjs
Shourn: I'd argue no.
For the most part `Document` and `PseudoDocument` should be interchangeable, it shouldn't matter if `parentDocument` is pseudo or not.
For the other cases, mostly related to database operations, I find being specific about requiring the `parentFoundryDocument` helps with understanding how the code works.

module/pipelines/effects.mjs
Shourn: I'm of the believe that irreversible operations **always** need confirmation.

module/projectfu.mjs
Shourn: This isn't even needed anymore. I'll just remove it.

templates/item/partials/item-armor-description.hbs
Shourn: Be aware, this is not the select for the type of slots.
These are the visible slots on the sheet.
If this were to move to the attributes tab, attributes would need to be made the default tab for user experience reasons.

module/documents/pseudo/pseudo-document-enabled-type-data-model.mjs
Shourn: This class could be _made_ into a mixin but it cannot just reuse the existing `EnablePseudoDocumentsMixin`.
Since `TypeDataModel` is basically a blank slate for extension by design I found it more natural to create a specialized subclass for this purpose.

module/documents/pseudo/pseudo-document.mjs
Azurelol: Ah, it was mistaken. Never mind.

module/documents/pseudo/pseudo-document.mjs
Azurelol: Very well.

templates/item/partials/item-armor-description.hbs
Azurelol: Aye. Then I will bring it up again at a later time.

PR #568, opened 2025-10-19
Invocation Selection Application Template Redesign

Unarekin:
The `InvocationSelectionApplication` layout is unintuitive, including enriched buttons that are not meant to be clicked, and unobvious buttons that *are* meant to be clicked.

The intent here is to mimic the various item tables on actor sheets to provide a more intuitive interface for outputting an Invocation to chat, as well as prevent the ability to click on the enriched buttons that will just throw an error anyway.

image

### DIFF
--- a/module/helpers/tables/invocation-table-renderer.mjs
+++ b/module/helpers/tables/invocation-table-renderer.mjs
@@ -56,7 +56,7 @@ export class InvocationTableRenderer extends FUTableRenderer {
 	}
 
 	static #renderNameHeader() {
-		return 'FU.Name';
+		return game.i18n.localize('FU.Name');
 	}
 
 	static async #renderName(invocation) {


module/helpers/tables/invocation-table-renderer.mjs
```diff
@@ -0,0 +1,65 @@
...
+	static #renderControlsHeader() {
+		return `<div class="header-item-controls"></div>`;
+	}
+
+	static #renderControls(invocation) {
+		return `<div class="cell-item-controls"><a class="cell-item-controls__control" data-action="roll" data-tooltip="${game.i18n.localize('FU.ChatMessageSendHint')}"><i class="fa-solid fa-share"></i></a></div>`;
+	}
+
+	static #renderNameHeader() {
+		return 'FU.Name';
```
Shourn: This is not automatically localized, you'll want to add a `game.i18n.localize` call.

Unarekin: Ope.  Fixed.

PR #580, opened 2025-10-24
bugfix: Actor and Item images getting cropped

Shourn:
- adjust how actor and item sheet header images are displayed to avoid cropping them


Azurelol (approved):
If this leads to an acceptable output for all, then I approve.

PR #581, opened 2025-10-25
bugfix: Ensure meta currency values are parsed as integers in update function

Bleck-42:
Manual modifications to the Metacurrency form cause it to incorrectly format as a string.

This change ensures values from form data are parsed as integers and default to zero if invalid.

### Repo Steps:
Use points
image

Change value from `2`
image


Shourn:
Root cause addressed in #583, many thanks for putting me on the right track!

PR #597, opened 2025-10-29
Crisis Score Modifiers

Unarekin:
- Implements a global Crisis score multiplier in the Homebrew Settings menu
- Standardizes several hbs templates' use of `lte` and `pfuHalf` vs. `pfuCrisis`
- Adjusts `pfuCrisis` handlebars helper to allow for passing an actor
- Adds `pfuCrisisPerc` handlebars helper
- Adds `system.resources.hp.crisis` value to CharacterDataModel and NPCDataModel
- Adds `system.resources.hp.crisis` to ActiveEffect attribute keys
- Adds `updateActiveEffect` hook to check for ActiveEffects that affect HP values (current, max, crisis) to re-check if the Crisis status effect needs to be added or removed from its target Actor

### DIFF
--- a/module/documents/effects/active-effect-config.mjs
+++ b/module/documents/effects/active-effect-config.mjs
@@ -203,7 +203,6 @@ function getAttributeKeys(document) {
 				attributeKeys.push(`system.resources.${res}.max`);
 			}
 			attributeKeys.push('system.resources.hp.crisisScore');
-			attributeKeys.push('system.resources.hp.inCrisis');
 			// Attributes
 			for (const attr of Object.keys(FU.attributes)) {
 				attributeKeys.push(`system.attributes.${attr}`);


module/documents/actors/npc/npc-data-model.mjs
```diff
@@ -193,6 +194,20 @@ export class NpcDataModel extends foundry.abstract.TypeDataModel {
 			},
 		});
 
+		Object.defineProperty(this.resources.hp, 'crisis', {
+			configurable: true,
+			enumerable: true,
+			get() {
+				const max = this.max;
+				const multiplier = game.settings.get(SYSTEM, SETTINGS.optionCrisisMultiplier) ?? 0.5;
```
Azurelol, line 201: Hmm? Isn't the default for the crisis multiplier be set to 0.5? As in the very first time this gets called it should properly return a value.

Unarekin: Yeah, the nullish operator is there in case there's some error getting the game setting, to ensure we aren't trying to multiply by `undefined`

module/documents/actors/actor.mjs
```diff
@@ -189,19 +189,13 @@ export class FUActor extends Actor {
...
-	async _onUpdate(changed, options, userId) {
+	async applyCrisis() {
 		if (this.isCharacterType) {
 			const { hp } = this.system?.resources || {};
-
-			if (hp && userId === game.userId) {
-				const crisisThreshold = Math.floor(hp.max / 2);
-				const shouldBeInCrisis = hp.value <= crisisThreshold;
+			if (hp) {
+				const shouldBeInCrisis = hp.value <= hp.crisis;
```
Shourn: Would it make sense to move this calculation into the data models?
Maybe defined on the HP resource object?
```javascript
get inCrisis() {
  return this.value <= this.crisis;
}
```
Although in that case I'd like the `crisis` field to be renamed to `crisisScore`.

module/helpers/handlebars.mjs
```diff
@@ -79,11 +82,27 @@ export const FUHandlebars = Object.freeze({
...
-			value = parseFloat(value);
-			max = parseFloat(max);
-			const half = max / 2;
-			return value <= half;
+		/**
+		 * Helper to determine if a given Actor is in Crisis.
+		 * When given an FUActor, it will respect the Actor's Crisis modifiers
+		 * When given two numeric values, it will respect the global Crisis modifier but NOT an Actor's specific modifier(s)
+		 */
+		Handlebars.registerHelper('pfuCrisis', function (...args) {
```
Shourn: This entire helper could be made obsolete by introducing an `inCrisis` getter as described in the comment on Actor.mjs.

templates/actor/partials/actor-point-bars.hbs
```diff
@@ -3,23 +3,23 @@
...
             {{/if}}
         </label>
     </div>
     <div class="bar bar-progress">
-        {{#if (pfuCrisis system.resources.hp.value system.resources.hp.max)}}
+        {{#if (pfuCrisis actor)}}
             <div class="bar-progress-bar bar-hp-crisis"
                  style="width: {{pfuPercentage system.resources.hp.value system.resources.hp.max}};"></div>
-            <div class="center-line"></div>
+            <div class="center-line" style="left:{{pfuCrisisPerc actor}}%"></div>
```
Shourn: You should be able to solve this without having to introduce a new helper by using the css calc function:
`left: calc({{actor.system.resources.hp.crisis}} / {{actor.system.resource.hp.max}} * 100%);`

Unarekin: My original thinking was that there may be some situation in the future where someone needs to refer to the value outside of placing the center line, but on retrospect it seems pretty unlikely and using `calc` here is a cleaner separation of display and data.

module/documents/actors/actor.mjs
Unarekin: Yeah sure, it encapsulates the actual logic and while I doubt the comparison will ever need to be changed, maybe it'll make things a little easier to reference if an actor is in Crisis elsewhere

There _are_ some parts of the system that specifically check if an actor has the `crisis` condition, and it may be worth revisiting those to see if they would want to shift to using the `inCrisis` getter.

Azurelol: I agree with renaming the field since `crisis` itself is more a state, which is best represented as a boolean.

Azurelol: Having slept on it, I think `crisisThreshold` for the field would suit best.

module/documents/actors/npc/npc-data-model.mjs
```diff
@@ -193,6 +194,20 @@ export class NpcDataModel extends foundry.abstract.TypeDataModel {
 			},
 		});
 
+		Object.defineProperty(this.resources.hp, 'crisis', {
```
Shourn: Could you please align this to how it is implemented in `CharacterDataModel`?

Shourn (changes requested):

module/documents/actors/character/character-data-model.mjs
```diff
@@ -203,6 +203,31 @@ export class CharacterDataModel extends foundry.abstract.TypeDataModel {
...
+		Object.defineProperty(this.resources.hp, 'inCrisis', {
+			configurable: true,
+			enumerable: true,
+			get() {
+				return this.value <= this.crisisScore;
+			},
+			set(newValue) {
+				delete this.inCrisis;
+				this.inCrisis = newValue;
+			},
```
Azurelol: Regarding this property, I think it would be more suitable to have it as part of the system data model, so `actor.system.crisis` would be the property to check.

Would that cause trouble for your current design?

module/documents/actors/npc/npc-data-model.mjs
```diff
@@ -193,6 +194,31 @@ export class NpcDataModel extends foundry.abstract.TypeDataModel {
...
+		Object.defineProperty(this.resources.hp, 'inCrisis', {
+			configurable: true,
+			enumerable: true,
+			get() {
+				return this.value <= this.crisisScore;
+			},
+			set(newValue) {
+				delete this.inCrisis;
+				this.inCrisis = newValue;
+			},
```
Azurelol: How come this properpy has a setter? Shouldn't it be read-only?

Unarekin: Ope, oversight on my part.

module/documents/actors/character/character-data-model.mjs
Unarekin: Honestly, I'm generally of the opinion that the crisis properties should be alongside the other HP-related properties, where they are now.

module/documents/effects/active-effect-config.mjs
```diff
@@ -202,6 +202,8 @@ function getAttributeKeys(document) {
 			for (const res of resources) {
 				attributeKeys.push(`system.resources.${res}.max`);
 			}
+			attributeKeys.push('system.resources.hp.crisisScore');
+			attributeKeys.push('system.resources.hp.inCrisis');
```
Azurelol: We probably want to remove this since it's only a getter as well.

module/documents/actors/character/character-data-model.mjs
Azurelol: Very well.

PR #598, opened 2025-11-01
bugfix: CombatHUD button issues

Bleck-42:
- Refactored drag position calculations to use initial mouse coordinates and element positions - Fixes the position jumping bug
  - Empty data check + try/catch - Fixes the JSON.parse crash
  - `_doDragOver` implementation with preventDefault() - Required for HTML5 drag/drop to work
  - Dragover event listener wiring - Required for drops to work
  - popout functionality was using jQuery methods that don't exist in ApplicationV2

### DIFF
--- a/module/ui/combat-hud.mjs
+++ b/module/ui/combat-hud.mjs
@@ -445,6 +445,7 @@ export class CombatHUD extends foundry.applications.api.HandlebarsApplicationMix
 		const hoverIn = this._onHoverIn.bind(this);
 		const hoverOut = this._onHoverOut.bind(this);
 		const dragCombatantStart = this._doCombatantDragStart.bind(this);
+		const dragCombatantOver = this._doDragOver.bind(this);
 		const dragCombatantDrop = this._doCombatantDrop.bind(this);
 
 		for (const row of rows) {
@@ -454,6 +455,7 @@ export class CombatHUD extends foundry.applications.api.HandlebarsApplicationMix
 
 				if (game.settings.get(SYSTEM, SETTINGS.optionCombatHudReordering) && game.user.isGM) {
 					row.addEventListener('dragstart', dragCombatantStart);
+					row.addEventListener('dragover', dragCombatantOver);
 					row.addEventListener('drop', dragCombatantDrop);
 				}
 			}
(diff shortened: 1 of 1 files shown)


module/ui/combat-hud.mjs
```diff
@@ -810,7 +818,18 @@ export class CombatHUD extends foundry.applications.api.HandlebarsApplicationMix
...
+		if (!rawData) return;
+
+		let data;
+		try {
+			data = JSON.parse(rawData);
+		} catch (error) {
+			console.warn('Combat HUD: Invalid JSON data in drop event', error);
+			return;
+		}
+
```
Unarekin, line 832: This is a good catch to ensure we aren't attempting to operate on malformed data, but I do also think we should be addressing situations where that data is set in the first place.

I don't believe any particular bugs have been reported about this, is this meant to address something in particular or is it more of a preventative measure?

module/ui/combat-hud.mjs
```diff
@@ -892,9 +911,12 @@ export class CombatHUD extends foundry.applications.api.HandlebarsApplicationMix
...
 			this._poppedOut = true;
-			this.element.find('.window-popout').css('display', 'none');
-			this.element.find('.window-compact').css('display', 'none');
-			this.element.find('.window-minimize').css('display', 'none');
+			const popoutButton = this.element.querySelector('.window-popout');
+			const compactButton = this.element.querySelector('.window-compact');
+			const minimizeButton = this.element.querySelector('.window-minimize');
+			if (popoutButton) popoutButton.style.display = 'none';
+			if (compactButton) compactButton.style.display = 'none';
+			if (minimizeButton) minimizeButton.style.display = 'none';
```
Unarekin, line 919: Ope, good catch.

module/ui/combat-hud.mjs
Bleck-42: I didn't see an issue reported, more things I found while using the hud and some fixes to patch things up.

I definitely echo the sentiment to fix it upstream. I don't think anything negative came from this, but it was throwing an error in the console logs on init. So technically yes it addresses something, but feels more preventative

Unarekin (approved):

PR #622, opened 2025-11-22
Combat HUD settings user scope

Unarekin:
v13 introduced a `user` scope for settings, separate from the `client` scope.
Many of the currently `client`-scoped system settings are either intended to be or are expected to be per *user* rather than per *device*.

This PR moves the combat HUD settings to the `user` scope, while setting their default values to the value of the same setting from the `client` scope, if set, or the usual static value.

image

### DIFF
--- a/module/settings.js
+++ b/module/settings.js
@@ -304,7 +304,7 @@ export const registerSystemSettings = async function () {
 		name: game.i18n.localize('FU.ExperimentalCombatHudSettings'),
 		hint: game.i18n.localize('FU.ExperimentalCombatHudSettingsHint'),
 		label: game.i18n.localize('FU.ExperimentalCombatHudSettingsLabel'),
-		scope: 'client',
+		scope: 'user',
 		icon: 'fas fa-book',
 		type: CombatHudSettings,
 	});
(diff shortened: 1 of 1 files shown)


module/settings.js
```diff
@@ -94,6 +94,29 @@ export const SETTINGS = Object.freeze({
...
 
+/**
+ * Attempts to retrieve a setting from the client scope specifically
+ * Used for the defaults for some user scope settings later, to migrate
+ * them to user scope.
+ * @param {string} setting - Key for the setting to retrieve
+ * @param {any} def - Default value to return
+ * @returns
+ */
+function getClientSetting(setting, def) {
```
Azurelol: I am a bit triggered by calling a param `def`. Let's use `default` or `defaultValue` here.

Azurelol (changes requested):

module/settings.js
Unarekin: ... what's wrong with `def`?
(also we can't call it `default` that's a keyword)

Azurelol: Sorry, that was rather too terse. (Long day); 

There are several languages where 'def' is a keyword for declaring variables, with one of them being one I am rather familiar with professionally, so this parameter out to me. I think it's worth being a bit verbose with param and variables.

Azurelol (approved):

PR #626, opened 2025-11-28
Party Sheet Fixes

ravynhardcastle:
This PR adjusts two things:
1. It makes the names in the party sheet not get cut off on letters like 'g' or 'y' _without_ removing the automatic ellipsis appending.
2. It makes the background 'colour' account for the level of the class, not just the class. 
 - The only difference in the logic other than that is that you are no longer considered 'any' (yellow) if one of your categories is equal to or greater than the other categories combined.
 - This PR does not adjust what categories each class is considered as, but I think with this new more flexible system it might be worth looking into? For instance, darkblade could be considered tank (although in this system having literally any levels in tank considers you a tank, so maybe thats too extreme for darkblade specifically,) symbolist, floralist, gourmet, and maybe commander could be considered support, and... pilot could be somewhere? It could arguably fit into any of the categories, really, so it's hard to say...
 - Perhaps this system should allow for classes to technically account for multiple categories, now that it is level-based? Perhaps certain weird classes shouldn't affect this at all, and we just hope the rest of their build defines it (currently classes are damage unless said otherwise)?
 - Either way, this PR addresses none of those questions, this just formalizes those questions to think about.

### DIFF
--- a/module/documents/actors/party/party-data-model.mjs
+++ b/module/documents/actors/party/party-data-model.mjs
@@ -192,6 +192,7 @@ export class PartyDataModel extends foundry.abstract.TypeDataModel {
 						name: item.name,
 						fuid: item.system.fuid,
 						img: item.img,
+						level: item.system.level.value,
 					};
 				});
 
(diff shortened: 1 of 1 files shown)


Azurelol (approved):
Neat!

PR #630, opened 2025-12-06
Fix default skill damage type to use weapon's type if damage from weapon is set to true

soujiroelric:
When a skill uses a weapon's data, it should use the weapon's type rather than defaulting to physical. Also, the default dash should communicate that the default weapon's type is being used.


templates/item/partials/item-damage-section.hbs
```diff
@@ -19,7 +19,7 @@
 			<label class="resource-label-m flexlarge">{{localize 'FU.DamageType'}}</label>
 			<select name="system.damage.type" class="resource-inputs select-dropdown-m" {{disabled (not system.damage.hasDamage)}}>
 				{{#if system.useWeapon.damage}}
-					{{selectOptions damageTypes selected=system.damage.type blank='-' sort=true localize=true}}
```
Azurelol, line 22: You will want to keep the selected property here, as that is the path of the property the selectOptions is working for.

For the blank, let's go with a localized succinct (single word) string. I think something like `Inherit` would work well here, as in "the type is inherited from the weapon". 

Since all our damage types are short single words, we would like to keep visual consistency.

You will want to add a new localized key in `en.json` by the end and use it here in a `{{localize KEY}}` block.

Finally, you will want to modify the default type in the `DamageDataModel` class, where it is actually set for new items.

soujiroelric: Addressed

Azurelol: Great.

module/documents/items/common/damage-data-model-v2.mjs
```diff
@@ -17,7 +17,7 @@ export class DamageDataModelV2 extends foundry.abstract.DataModel {
 			value: new NumberField({ initial: 0, integer: true, nullable: false }),
 			onRoll: new StringField({ blank: true }),
 			onApply: new StringField({ blank: true }),
-			type: new StringField({ initial: 'physical', choices: Object.keys(FU.damageTypes), blank: true, nullable: false }),
+			type: new StringField({ initial: '', choices: Object.keys(FU.damageTypes), blank: true, nullable: false }),
```
Azurelol, line 21: Now, rather than changing the default here (which would affect spells, among other things), we only want to assign it on the schema for the skill data model.

Like so (in skill data model class):

Image

PR #754, opened 2026-02-06
Update Compendium Entries for Esper Through Merchant

Frozedon:
This pull request is Compendium Changes and Updates to Skills and Features of the Esper, Floralist, Fury, Gourmet, Guardian, Invoker, Loremaster, and Merchant classes. Also, an update to Hydra Dance on Dancer. It tries to incorporate up-to-date automation while also cleaning up descriptions and messaging.

### General Summary
- Removed bulleted lists from inline effects in descriptions.
- Set Target Rule to "Self" for various Skills where targeting isn't useful.
- Attempted to add automation where it was possible.
- Any Skills that had targeting primarily on targets other than Self, moved Effect into the Attributes of the Skill.

## Detailed Breakdown
This section provides a detailed breakdown of the changes made to each Skill. Organized by Class.

### Dancer
- Hydra Dance. Replaced the 2 Update Resources with a single message that has description on the options and has inline buttons for the recovery.

### Esper
- Psychic Gifts. Cleaned up text.
- Hypercognition. Updated Rule Element to use the "Calculate Expenses" Rule Trigger. 

### Floralist
- Magiseeds. Cleaned up text. Also, added scaling damage and HP/MP recovery based on character level.
- Hookleaf Shade. Updated automation with rule elements.
- Wardwattle. Added Effects to notify when an enemy hits the floralist with a melee attack.
- Verdant Sway. Added Perform Ritual button through Rule Element.

### Fury
- Indomitable Spirit. Made options into bulleted list for readability.
- Provoke. Moved effects to Attributes.
- Withstand. Update inline buttons to use the highest bond strength variable for HP.

### Gourmet
- Traveling Cook. Added link to Ingredients Roll Table to quickly generate new ingredients.

### Guardian
- General clean up.

### Invoker
- Invocations. Added "You Spend" text and inline buttons to all the Invocation entries since no expenses show up normally for using the feature. Also, added scaling by character level. Separated out inline buttons to use when Elemental Harmony or Wellspring Expansion should be used.

### Loremaster
- Flash of Insight. Added Rule Element that displays a message to notify about asking questions when the loremaster makes an INS+INS Open Check and the result is 13+.
- Focused. Set Accuracy to INS+INS and added the SL as a bonus.

### Merchant
- General clean up.


Azurelol (approved):
Great work.

PR #785, opened 2026-02-10
Dice So Nice Doubles Effect Fix

### LINKED ISSUE #663: 4.5.0-rc.6 Dice So Nice! Integration Broken
Project FU: 4.5.0-rc.6
Dice So Nice! 5.2.4

**Bug Description:**
With the dice so nice special effects integration turned on dice will fail to roll. With the default behavior of not showing messages in chat until dice are finished rolling this means messages will never appear in chat when rolling items

**Steps to Reproduce:**
1. Activate Dice So Nice!
2. Navigate Configure Settings > Dice So Nice! > 3D Dice Settings > Special Effects
3. Add an effect with the trigger "Doubles"
4. Roll an item and observe nothing happening

**Expected behavior:**
Previously this integration would allow triggering special effects on double rolls per Fabula's Crit & Fumble rules.

Unarekin:
Doubles sfx in `diceSoNiceRollStart` hook was using incorrect roll result property, resulting in an error being thrown when effect is enabled

Fixes #663

### DIFF
--- a/module/projectfu.mjs
+++ b/module/projectfu.mjs
@@ -547,8 +547,8 @@ Hooks.once('diceSoNiceReady', (dice3d) => {
 	Hooks.on('diceSoNiceRollStart', (_messageId, context) => {
 		const dice = context.roll.dice;
 		if (dice.reduce((agg, curr) => agg + curr.number, 0) === 2) {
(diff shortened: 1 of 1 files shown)


Azurelol (approved):
Thanks for looking into this.

PR #787, opened 2026-02-10
Double Pressure With High Damage Threshold

### LINKED ISSUE #773: Pressure increase not accounting for high damage
**Bug Description:**
As per the playtest rules on Pressure, when a foe takes damage of a type they are vulnerable, they gain 1 Pressure. However there is one extra part that is not being accounted for. When said damage is equal to 10 + half the enemy's level, the pressure clock fills up by 2, rather than 1. This part of the pressure system is not being accounted for.
```
When a non-staggered enemy loses Hit Points due to damage of a type they are
Vulnerable to, that enemy does not lose twice the normal amount of Hit Points as
described on page 92 of the Core Rulebook; instead, their Pressure Clock fills by 1
section, or by 2 sections if the HP loss they suffered is equal to or higher than 【10
+ half that enemy's level】. For instance, the Clock of a level 25 enemy will fill by 2
sections if they lose 22 or more Hit Points from the blow.
```

**Expected behavior:**
Ideally this aspect of pressure clocks is correctly accounted for, since it's very important for builds that deal singular high damage bursts.

Unarekin:
This PR adds a check for the pressure system's `processVulnerability` function to increment an actor's Pressure clock by two steps when being triggered because of a damage type VU and the actual damage amount is >= 10 + (half the NPC's level)

Fixes #773 

image

### DIFF
--- a/module/systems/pressure-system.mjs
+++ b/module/systems/pressure-system.mjs
@@ -29,8 +29,19 @@ async function processVulnerability(context) {
 		return null;
 	}
 
+	console.log('Processing vulnerability:', context);
+
 	// TODO: Refactor to not have to re-resolve
-	pressureClock = await context.actor.updateProgress('pressure', 1);
+
+	// If this NPC is pressured because they are VU to the damage type,
+	// AND the amount of HP loss is equal to or higher than 10 + half their level,
+	// fill the clock by 2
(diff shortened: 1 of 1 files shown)


Azurelol (approved):
Thanks!

Azurelol (approved):

PR #796, opened 2026-02-12
Additional Dice So Nice! Triggers

### LINKED ISSUE #788: Dice So Nice Effects Triggers
The current method of adding special effects to rolls with Dice So Nice! when rolling a critical success or a fumble involves far more clicks that it really necessitates, and is unable to properly account for effects that alter a character's critical threshold, such as the Frenzy skill

Image

I propose we add:
- A Critical and Fumble effect trigger, which reference the ChatMessage's `flags.projectfu.CheckV2.critical` and `flags.projectfu.CheckV2.fumble` properties, respectively
- Possibly hit/miss and/or success/failure triggers for non-crit successes on Checks?

Unarekin:
This PR adds the following triggers for Dice So Nice! effects:
- Critical
- Fumble
- Success
- Failure

The Critical and Fumble triggers use the `critical` and `fumble` properties from the roll's ChatMessage CheckV2 data, which respects things that may expand critical threshold, like the Frenzy skill.

The Success and Failure triggers will trigger for all Check types:  Accuracy, Magic, Attribute, and Group Checks.
They will only trigger for the _second_ roll in an Opposed Check, as we cannot yet know success or failure on the first roll.

The original Doubles trigger is preserved for backwards compatibility.

image

### DIFF
--- a/module/projectfu.mjs
+++ b/module/projectfu.mjs
@@ -544,6 +544,8 @@ Hooks.once('diceSoNiceReady', (dice3d) => {
 		Array.from({ length: 20 }, (v, i) => (i + 1).toString()),
 	);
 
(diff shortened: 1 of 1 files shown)


Azurelol (approved):
Great work!

PR #812, opened 2026-02-16
Compendium Updates to Psychic Gifts

Frozedon:
Added inline buttons for spending MP and advancing the Brainwave Clock back to Psychic Gifts where they were removed.

In situations where a Player can't or won't use the Rule Element automation for any of the Psychic Gifts, It is useful to have the inline buttons to spend MP based off the Brainwave Clock filled segment and the button to automatically progress the clock in the description of every Psychic Gift.

Also, decided to update some of the labeling and add a direct link to the Psychic Gift List to the Psychic Gift Skill.


Azurelol (approved):
I agree with this change.

Issue #786, opened 2026-02-10
Players Cannot Loot From Party Inventory


Frozedon:
**Bug Description:**
While the Players have Observer permissions for the Party sheet, they are seeing the wrong button for looting inventory items and it does not loot to the player's inventory.

**Steps to Reproduce:**
1. Create a Player User.
2. Create a PC and give ownership of that PC to the Player.
3. Create a Party.
4. Give "Observer" permission to the Player.
5. Place an item into the Party Inventory.
6. Have the Player open the Party Sheet.
7. Have the Player attempt to loot the item in the Party Inventory.
8. Observe that the button is wrong and it doesn't do anything.

**Expected behavior:**
Players with Observer permission to the party should see a "Hand" icon button for looting, and pressing it should give the Player's PC that item and remove the item from the Party Inventory.

**Attachments (if applicable):**

<img width="915" height="934" alt="Image" src="https://github.com/user-attachments/assets/9bdd722c-05a5-4785-af32-f5192da84c8f" />

<img width="709" height="106" alt="Image" src="https://github.com/user-attachments/assets/14bb60f3-d0c8-4808-b69f-395e68191e09" />

Frozedon:
Fixed by #814

PR #813, opened 2026-02-16
Add Loot Item to Party Sheet for Players

Frozedon:
Created a fix for #786 .

The Party sheet only showed the "Share Item" button for both GMs and Players. But, that button automatically fails if the user does not own the Party Sheet which Players are not intended to own. This updates to match the old functionality where it would show a "Loot Item" button instead to Players.

- "Loot Item" Button appears on the Party sheet for Players. "Share Item" Button for GMs.
- Added Loot Item function to party sheet.


Azurelol:
While I am happy to see a bug fix contribution, in this case it's best to tackle the root cause and solve the issue across the board. Please take a look over #814 .

PR #900, opened 2026-03-14
Pressure Gauge and Sprite Scale

### LINKED ISSUE #807: Pressure Gauge and Sprite Scale
**Bug Description:**
Scaling a sprite can cause it to overlap with the pressure gauge, potentially hiding it from view entirely.

**Steps to Reproduce:**
1. Enable the Pressure system and Show Pressure Gauge options
2. Create an actor with a rank of Elite or Champion
3. Add their token to a scene, and set its scale (*not* dimensions) to something greater than 1
4. Profit

**Expected behavior:**
The pressure gauge should be visible at all times when it is enabled and pertinent

Image

Unarekin:
Adds handling for sprite scales greater than 1 to ensure pressure gauge remains visible

image

image

image

### DIFF
--- a/module/ui/pressureGauges/pressure-gauge.mjs
+++ b/module/ui/pressureGauges/pressure-gauge.mjs
@@ -106,7 +106,7 @@ export class FUPressureGauge extends globalThis.PIXI.Container {
 				break;
 			default:
 				if (scaleGauge) {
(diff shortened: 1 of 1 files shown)


Azurelol (approved):
Great!

PR #901, opened 2026-03-14
Pressure Clock as Trackable Resource

Unarekin:
This adds the pressure clock as a trackable resource for NPCs under `clocks.pressure`

[Pressure Resource.webm](https://github.com/user-attachments/assets/dbafacdc-6f09-4d10-b5a3-42588735a4ca)

Which also means:  Bar Brawl support.

[Pressure Resource - Bar Brawl.webm](https://github.com/user-attachments/assets/39b96e8f-4a84-4299-96a5-6246702f8f79)

### DIFF
--- a/module/documents/actors/npc/npc-data-model.mjs
+++ b/module/documents/actors/npc/npc-data-model.mjs
@@ -115,7 +115,6 @@ export class NpcDataModel extends BaseCharacterDataModel {
 			pressurePoints: new EmbeddedDataField(TraitsDataModel, {
 				options: TraitUtils.getOptionsFromConfig(FU.weaponCategories),
 			}),
-			clocks: new SchemaField({}),
 		});
 	}
 
@@ -131,7 +130,9 @@ export class NpcDataModel extends BaseCharacterDataModel {
 		this.#prepareReplacedSoldiers();
 		this.#prepareBasicResource();
 		this.derived.prepareData();
-		this.#preparePressureResource();
+
+		this.clocks = {};
+		this.#prepareClockResources();
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


module/documents/actors/npc/npc-data-model.mjs
```diff
@@ -130,6 +130,9 @@ export class NpcDataModel extends BaseCharacterDataModel {
 		this.#prepareReplacedSoldiers();
 		this.#prepareBasicResource();
 		this.derived.prepareData();
+
+		this.clocks = {};
```
Azurelol, line 134: To keep it consistent with the other rest of the API, I would like this to be named `tracks`. Since I consider tracks to be a more agnostic term that can be used for "progress tracks" (clocks) and "resource tracks" (resource bars).

Azurelol: And we can add this to the base character model  instead so it can be used to track certain PC class features as well. (If users feel like it).

module/documents/actors/npc/npc-data-model.mjs
```diff
@@ -221,6 +224,42 @@ export class NpcDataModel extends BaseCharacterDataModel {
 			},
 		});
 	}
+
+	/**
+	 * Adds a trackable resource to the clocks property of this DataModel
+	 * @param {string} key - Key name for the clock to be added to the DataModel
+	 * @param {string} fuid - fuid of the progress clock -- will be resolved with {@link FUActor.resolveProgress}
+	 */
+	addClockResource(key, fuid) {
```
Azurelol, line 233: Similarly here, `addTrackResource`. or `trackProgress`.

Azurelol (changes requested):

module/documents/actors/npc/npc-data-model.mjs
Unarekin: > To keep it consistent with the other rest of the API, I would like this to be named `tracks`. Since I consider tracks to be a more agnostic term that can be used for "progress tracks" (clocks) and "resource tracks" (resource bars).

What would fall under a "resource track" in this?

I chose the term `clocks` particularly because that is the name of the mechanic, and I did not anticipate setting up some kind of generic catch-all for any sort of changing value, but a place to store common clocks.

Azurelol: Anything that references the `ProgressDataModel` used in actors. That said, it's not a hill I would die on, and it can always be aliased.

Azurelol: That said, this declaration could be moved onto prepareClockResources (or removed altogether since you do null coalescing operation there).

Azurelol (approved):

PR #926, opened 2026-03-21
Generalize apex attributes implementation to support multiple out-of-bounds attributes with arbitrary values.

Braxbro:
Assuming I haven't made any dumb JavaScript errors, this should add the following Active Effects aspects:

attributes[attr].unlimit [bool]: If the base attribute die is at the max/min die size, uses abnormal if it is above/below max/min die size, respectively.
attributes[attr].abnormal [number]: The die size to be used when unlimit is true and base is the appropriate value.

This replaces the previous apex attribute implementation (I've yet to do the compendium changes for this. Obviously before the pull request is completed I intend to have the compendium changes finished.)

### DIFF
--- a/module/documents/actors/common/attribute-data-model.mjs
+++ b/module/documents/actors/common/attribute-data-model.mjs
@@ -32,7 +32,6 @@ export class AttributeDataModel extends foundry.abstract.DataModel {
 
 		// Set the initial current to start off the base value
 		this._current = this.base;
-		// Set the initial unlimit flag to start false
 		// Set the initial abnormal die size to d20 (Neo-Human)
 		this._abnormal = 20;
 
--- a/module/documents/actors/common/attribute-data-model.mjs
+++ b/module/documents/actors/common/attribute-data-model.mjs
@@ -8,20 +8,22 @@ function isEven(number) {
 	return number % 2 === 0;
 }
 
-// TODO: Provide support for NeoHuman
 const minimumValue = 6;
 const maximumValue = 12;
 
 /**
  * @property {number} base
  * @property {number} current
+ * @property {boolean} unlimit
+ * @property {number} abnormal
  * @property {number} bonus
  */
 export class AttributeDataModel extends foundry.abstract.DataModel {
 	static defineSchema() {
-		const { NumberField } = foundry.data.fields;
+		const { NumberField, BooleanField } = foundry.data.fields;
 		return {
 			base: new NumberField({ initial: 8, min: minimumValue, max: maximumValue, integer: true, nullable: false, validate: isEven }),
+			unlimit: new BooleanField({ initial: false, nullable: false })
 		};
 	}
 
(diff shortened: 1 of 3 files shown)


module/documents/actors/common/attribute-data-model.mjs
```diff
@@ -8,20 +8,22 @@ function isEven(number) {
...
 
-// TODO: Provide support for NeoHuman
 const minimumValue = 6;
 const maximumValue = 12;
 
 /**
  * @property {number} base
  * @property {number} current
+ * @property {boolean} unlimit
+ * @property {number} abnormal
```
Azurelol, line 18: I find these two property names are rather unintuitive, abnormal in particular. I would explore other alternatives here.

Azurelol (changes requested):
As an new contributor, part of a behavior-altering submission should include the "why" of the feature to be added.  In this case it's clearly been submitted to support your homebrew designs, which is understandable.

That said, I would also like to see a different approach here that doesn't force the compendiums to update. The existing (simplistic) implementation of setting a flag to override the die was to support one of the official designs of the system that many users asked for. 

When those compendium entries got published it became a contract to the users of the system. One that shouldn't be easily broken unless there's a very good reason to do so. If this solution were to be deployed, how many users do you see wanting to make use of this particular extension? If you were to sit on this for a month or two, would you still want to have it or would you have found a different way to accomplish your designs?

I would be more accepting of an approach that doesn't break the current apex attribute support. Yes, this solution is clearly more extensible, but it also affects one of the core data models of the system.

On top of the requested  and any future changes you may make, I would ask of you to set your expectations accordingly as I will have to chew on this for a while.

Unarekin:
Another thing to note:  As this is intended to replace the current apex attribute implementation, it is a *breaking* change, which will necessitate a major release.

Foundry has stated they only plan to have one more test release of Foundry v14 before it goes stable, which means we already have a major release in the very near future when that happens.

I would not be in favor of a major release between now and then

module/documents/actors/common/attribute-data-model.mjs
Braxbro: > I find these two property names are rather unintuitive, abnormal in particular. I would explore other alternatives here.

I see. I was considering calling it abnormal_size or something but wasn't sure if two-word property names were against convention.

Braxbro:
> As an new contributor, part of a behavior-altering submission should include the "why" of the feature to be added. In this case it's clearly been submitted to support your homebrew designs, which is understandable.
> 
> That said, I would also like to see a different approach here that doesn't force the compendiums to update. The existing (simplistic) implementation of setting a flag to override the die was to support one of the official designs of the system that many users asked for.
> 
> When those compendium entries got published it became a contract to the users of the system. One that shouldn't be easily broken unless there's a very good reason to do so. If this solution were to be deployed, how many users do you see wanting to make use of this particular extension? If you were to sit on this for a month or two, would you still want to have it or would you have found a different way to accomplish your designs?
> 
> I would be more accepting of an approach that doesn't break the current apex attribute support. Yes, this solution is clearly more extensible, but it also affects one of the core data models of the system.
> 
> On top of the requested and any future changes you may make, I would ask of you to set your expectations accordingly as I will have to chew on this for a while.

Noted - I will undo the commit breaking the apex integration and instead redirect it to be synonymous with setting the unlimit flag for the stat. The die size that enables defaults to 20 (or should, if I forgot it) so it should be equivalent to the current implementation.

PR #956, opened 2026-03-29
fix: clipping buttons and missing chat background

Re4XN:
* Fixes clipping at bottom of buttons in many tabs
  * Padding is probably not the perfect solution, but given how the styling of the buttons and the differences between tabs, it's probably the most straightforward / generic one.
* Fixes incorrect path for chat background


Azurelol (approved):

Azurelol:
Thank you for your submission!

PR #961, opened 2026-03-30
Compendium - Sample Rare Items & Effects

Frozedon:
Added a series of sample rare items that can be easily automated through the system. Additionally, added Equipment and Actor Effects for many common uses of automation in the system.

- **"Sample Rare Equipment" Compendium Added**. Filled with various equipment that use automation available in the system. The items aren't exhaustive, but try to be examples of many common applications. The equipment is organized by equipment type and then folders to label the Quality group that they have.
- **"Equipment Effects" Compendium Updated**. Added a series of effects that cover any of the use cases found in the "Sample Rare Equipment" along with many other common applications of automation. This tries to be exhaustive, with an effect for each use case. Like Affinities for each damage type, attribute changes for each attribute, immunities for each status effect, etc. These effects are meant to be applied to rare equipment created by a GM or Player, by dropping them directly onto the equipment item's effects. They are all made as Passive Effects.
- **"Actor Effects" Compendium Updated**. Any of the effects added to the "Equipment Effects" has their own version here, along with other common effects. These effects are meant to be applied directly to a PC or NPC. They are all made as Temporary Effects with a Duration of "None", intending for the GM or Player to handle application and removal.
- **Color Coding**. All the above Compendiums have been updated with a color coding to their folders for ease of use and visual organization.
- **Potion Fix**. The Remedy and Elixir Consumables were missing the "HitPoint" and "MindPoint" Traits respectively.


Unarekin:
Should probably mark this one as a draft until we get clarification on the licensing terms for those images, lest someone accidentally merge it in

spyrella:
> Should probably mark this one as a draft until we get clarification on the licensing terms for those images, lest someone accidentally merge it in

Since pixeltier assets were used for example rare items, looks like the assets are covered by the license.

spyrella (approved):

Azurelol (approved):
I express some reservations regarding being this exhaustive regarding character options in the compendium but hopefully I will be proven wrong.

PR #968, opened 2026-03-31
More Clock Trackable Resources

Unarekin:
- Moves `addClockResource` function from `NPCDataModel` to `BaseDataModel`, which it should have been in in the first place (and I just completely missed azurelol's request to move it in the previous PR, woops)
- Only registers the pressure clock as a trackable resource if the pressure system is enabled
- Adds Zero Power clock as a trackable resource, if Zero Powers are enabled
  - This expects the Zero Power to have the fuid `zero-power`, which is the default when creating one
- Adds the Brainwave Clock as a trackable resource, using the fuid `brainwave-clock`
- Adds Garden Clock as a trackable resource, using the fuid `floralist-garden`

### DIFF
--- a/module/documents/actors/character/character-data-model.mjs
+++ b/module/documents/actors/character/character-data-model.mjs
@@ -240,7 +240,7 @@ export class CharacterDataModel extends BaseCharacterDataModel {
 
 	#prepareClockResources() {
 		this.addClockResource('brainwave', 'brainwave-clock');
-		this.addClockResource('garden', 'floralist-garden');
+		this.addClockResource('garden', 'garden');
 		this.addClockResource('zeroPower', 'zero-power');
 	}
 }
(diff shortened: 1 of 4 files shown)


module/documents/actors/character/character-data-model.mjs
```diff
@@ -236,4 +237,10 @@ export class CharacterDataModel extends BaseCharacterDataModel {
 		this.resources.fp.value = this.resources.fp.value || 0;
 		this.resources.exp.value = this.resources.exp.value || 0;
 	}
+
+	#prepareClockResources() {
+		this.addClockResource('brainwave', 'brainwave-clock');
+		this.addClockResource('garden', 'floralist-garden');
```
Azurelol: This should be `garden` now, as I reverted the id change after our discussion.

Azurelol (changes requested):

module/documents/actors/character/character-data-model.mjs
Unarekin: Hrm, maybe I hadn't updated after all, will fix

Azurelol (approved):

PR #1012, opened 2026-04-11
Actor sheet stats page layout overhaul

Element-Re:
Updates the stats page on PC and NPC sheets to use an improved flex layout. Limits scrolling needed and keeps relevant information in focus.


Azurelol (approved):
Great work.

PR #1014, opened 2026-04-11
Compendium draggable clarity

4m2c4bl3:
Cursor only changes to a draggable indicator on a compendium item, when the user has their cursor over the image.


Azurelol (approved):
Thank you very much!

PR #1040, opened 2026-04-14
Compendium - Class Attributes Now Use fu-ids

Frozedon:
Updated all Skills, Spells, and Heroic Skills to use the fu-ids of Classes in their Attributes. That way they can properly reference the correct Classes without breaking when the Class name is changed.

Also, some minor text edits to remove unused ascii characters in the spacing of descriptions.


Azurelol (approved):
Thank you for your work. It's greatly appreciated.

PR #1056, opened 2026-04-16
Add and improve layout options for actor sheet front page

Element-Re:
An overhaul of the actor sheet front page to better support both single-column and multi-column layouts as options. 

As discussed with @Azurelol, while this includes the CSS and template changes to support the ability to present both single and multi-column layouts, the layout switch to multi-column mode is based on the presence of the `multi-column` class on the front page `.body-section` element, and this PR does not include any logic for adding that class to the rendered sheet.

Some additional clock styling improvements are also included.


Azurelol (approved):
Thank for your work.

PR #1058, opened 2026-04-16
Update NPC Profile to support pressure points

### LINKED ISSUE #774: Enemy Study page not including Pressure Points
**Bug Description:**
Kinda what it says on the tin. Pressure points are not being included in alongside other affinities when a target is studied. I feel like they should be listed right next to affinities. (Also i noticed the study page lacks a scroll bar, which means you can't reach the end of some records) 

Image

Azurelol:
- Update NPC profile edit dialog
- Update NPC profile view
- Add support for pressure points on NPC profile, Limited NPC sheet
- Reveal pressure point information on damage

image

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1776,6 +1776,7 @@
 		"PressureClock": "Pressure Clock",
 		"PressurePoints": "Pressure Points",
 		"PressurePoint": "Pressure Point",
+		"PressurePointAbbr": "PP",
 		"Stagger": "Stagger",
 		"Compendium": "Compendium",
 		"CompendiumBrowser": "Compendium Browser",
--- a/module/helpers/handlebars.mjs
+++ b/module/helpers/handlebars.mjs
@@ -264,6 +264,7 @@ export const FUHandlebars = Object.freeze({
 		Handlebars.registerHelper('pfuTraits', traits);
 		Handlebars.registerHelper('pfuIconAttribute', attributeIcon);
 		Handlebars.registerHelper('pfuBadge', badge);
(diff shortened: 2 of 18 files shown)


Unarekin (changes requested):
The pressure point list does not show if there are no affinities revealed, whether through Studying or them being individually discovered via damage or editing the adversary profile

Image

Image

Azurelol:
> The pressure point list does not show if there are no affinities revealed, whether through Studying or them being individually discovered via damage or editing the adversary profile
> Image Image

Oh, I see the problem. I tied their reveal status together. I will iterate.

Unarekin:
> > The pressure point list does not show if there are no affinities revealed, whether through Studying or them being individually discovered via damage or editing the adversary profile
> > Image Image
> 
> Oh, I see the problem. I tied their reveal status together. I will iterate.

To be fair it seems incredibly unlikely that someone would be in a position where they would know a pressure point without also knowing a damage affinity, but it's technically possible.

Azurelol:
> > > The pressure point list does not show if there are no affinities revealed, whether through Studying or them being individually discovered via damage or editing the adversary profile
> > > Image Image
> > 
> > 
> > Oh, I see the problem. I tied their reveal status together. I will iterate.
> 
> To be fair it seems incredibly unlikely that someone would be in a position where they would know a pressure point without also knowing a damage affinity, but it's technically possible.

Indeed. It's an edge case but easy enough to address.

Azurelol:
It should be addressed now.

Unarekin (approved):

PR #1059, opened 2026-04-17
More layout adjusments

Azurelol:
- Fix up track rendering on party sheet
- Update effect shortcuts
- Add bonds section option for PC sheet
- Hide favorites from NPCs when not used
- Hide biography and notes on NPCs when its tab is enabled
- Add sheet option to open the PFU specific sheet settings
- Fix progress bar height on actor sheets
- Fix up multi column layout when main section is not rendered (still needs work)
- Add a decorative fill on multi column layout

### DIFF
--- a/module/settings.js
+++ b/module/settings.js
@@ -82,6 +82,7 @@ export const SETTINGS = Object.freeze({
 	showAssociatedTherioforms: 'showAssociatedTherioforms',
 	showAdvancementTracker: 'showAdvancementTracker',
 	optionNPCNotesTab: 'optionNPCNotesTab',
+	optionPCBondsSection: 'optionPCBondsSection',
 	optionAlwaysFavorite: 'optionAlwaysFavorite',
 	optionAutomaticAdversaryRegistration: 'optionRegisterAdversaries',
 	partySheetTheme: 'optionPartySheetTheme',
@@ -170,6 +171,7 @@ export const registerSystemSettings = async function () {
 		icon: 'fas fa-book',
 		type: createConfigurationApp('FU.SheetOptions', [
 			SETTINGS.optionNPCNotesTab,
+			SETTINGS.optionPCBondsSection,
 			SETTINGS.optionAlwaysFavorite,
 			SETTINGS.showAssociatedTherioforms,
 			SETTINGS.showAdvancementTracker,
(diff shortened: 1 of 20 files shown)


module/settings.js
```diff
@@ -207,6 +209,15 @@ export const registerSystemSettings = async function () {
 		requiresReload: true,
 	});
 
+	game.settings.register(SYSTEM, SETTINGS.optionPCBondsSection, {
+		name: game.i18n.localize('FU.BondsSectionSettings'),
+		scope: 'world',
+		config: false,
+		type: Boolean,
+		default: false,
```
Unarekin, line 217: I do think this option should default to true.  I am generally of the opinion that when introducing the option to toggle something on or off, the default should maintain the way it was prior to that option being introduced.

I don't think the poll we ran on the discord is necessarily a definitive indicator of general public opinion on the presence of the bond section, owing to its short duration and the way the options were worded.

Unarekin (changes requested):

module/settings.js
Azurelol: I disagree on being tied down to previous design choices on something non-critical, particularly if we provide users a way to bring it back. 

If you would like to see another poll to confirm public sentiment then we can collaborate on the wording on the channel.

Unarekin: I can agree with that

styles/css/global/sheet.css
```diff
@@ -380,22 +380,32 @@
...
-						max-width: unset;
+					.section-container {
+						min-width: 170px;
 					}
 				}
-				.sidebar-container {
-					flex-wrap: wrap;
-					width: auto;
+
+				.fill {
```
Unarekin, line 403: I'm unsure how I feel about having a background on this fill element.  On most themes it looks alright, but on Reactor Five and Fabula Knights it looks a bit out of place.

Image

Image

Unarekin (dismissed):

styles/css/global/sheet.css
Azurelol: Ah, indeed. It doesn't look good on all themes (I tested) ad this was the best compromise I could make on limited time. 

I am not attached to the current solution: did you have other ideas for managing the fill so it looks decent across all themes?

Azurelol:
Oh, I am sorry about that.  I didn't see your last message.

Thank you for having taken a look at it.

Azurelol:
If the default option for the bonds section is holding this PR back from being merged it can be changed to whatever default you would like as it's now your prerogative. (I can do so now or you may after merging).

spyrella:
> If the default option for the bonds section is holding this PR back from being merged it can be changed to whatever default you would like as it's now your prerogative. (I can do so now or you may after merging).

My personal opinion on it is I'm fine with everything else besides decorative fill and opt in to show bond (could be opt out instead). Mainly until a better system for hiding and showing sections in this tab is implemented.

Azurelol:
Very well. I will re-open the PR from my fork and resubmit.

PR #1068, opened 2026-04-18
change zenit and rest groups to under the portrait

Re4XN:
image

* Moves Zenit group under the portrait
* Moves Rest button under the portrait
* Replaces inconsistent tab icons with text to be in line with rest of tabs
* Makes portrait 200x200


spyrella:
This is good, once the poll is finalized the decision will be made whether to add it. Thanks for your contribution.

Azurelol (approved):
Great work. As DMG said, once the consensus is gathered we can move ahead with this.

PR #1071, opened 2026-04-19
Add context menu option on actor sheets to send items to chat

### LINKED ISSUE #1062: Send to Chat Context Menu Entry
**User Story:**
As a player/gm, being able to send an item to the chat to show its description/stats/etc. without actually rolling the item and triggering any side effects is incredibly useful, but the option is tucked away in the kebab menu on the item sheet.

Image

Adding this option to the context menu on the actor sheet would help access this feature more quickly

Image

Azurelol:
image

### DIFF
--- a/module/sheets/actor-sheet-utils.mjs
+++ b/module/sheets/actor-sheet-utils.mjs
@@ -271,7 +271,7 @@ function getItemFromHtml(htmlElement, actor) {
  * @returns {Promise<void>}
  */
 async function _onItemDuplicate(element, sheet) {
-	let item = getItemFromHtml(element, sheet.actor);
+	const item = getItemFromHtml(element, sheet.actor);
 	if (item) {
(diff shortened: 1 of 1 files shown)


spyrella:
Wrong description  `* Duplicates the specified item and adds it to the actor's item list.`

it should be ` * Sends the specified item to chat via display check.`

Also, maybe `let item = getItemFromHtml(element, sheet.actor);` should be a `const` since it's only being referenced once and not reassigned.

Azurelol:
> Wrong description `* Duplicates the specified item and adds it to the actor's item list.`
> 
> it should be ` * Sends the specified item to chat via display check.`
> 
> Also, maybe `let item = getItemFromHtml(element, sheet.actor);` should be a `const` since it's only being referenced once and not reassigned.

Ack; I copy pasted the previous one and forgot to update the description. And of course whenever possible.

PR #1075, opened 2026-04-20
Drain Spirit + $hr Expression + Resource Pipeline Updates

Frozedon:
This work is all about updating the **Drain Spirit** spell to use the Resource Pipeline instead of the Damage Pipeline. It involved adding a new Expression "$hr", updating the Resource Pipeline to use Traits, and updating the "Calculate Resource" Rule Trigger to work for **Harvester of Sorrow**. 

- Added $hr expression. If it's part of a check, it is replaced with the High Roll of the Check.
- Fixed Resource Evaluation to include the check as part of the context.
- Fixed Magic Check to add Resource if set.
- Updated the Drain Spirit spell to use the Resource Pipeline instead of the Damage Pipeline.
- Added Traits to the Resource Pipeline.
- Made the Resource Pipeline be able to handle Absorb from Traits.
- Added "HitPointAbsorption" Trait. Only need to check in Resource Pipeline, since we can assume damage is HP by default.
- Updated Harvester of Sorrows Rule Element to alter the traits and resource amount of Drain Spirit.
- Added Resource Source, Local, and Identifier options to the Calculate Resource Rule Trigger.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -91,6 +91,7 @@
 		"Resource": "Resource",
 		"ResourcePoints": "Resource Points",
 		"ResourceStatus": "Resource Points?",
+		"ResourceSource": "Resource Source",
 		"Initiative": "Initiative",
 		"InitiativeCheck": "Initiative Check",
 		"InitiativeAbbr": "INIT",
--- a/lang/en.json
+++ b/lang/en.json
@@ -91,7 +91,6 @@
 		"Resource": "Resource",
 		"ResourcePoints": "Resource Points",
 		"ResourceStatus": "Resource Points?",
-		"ResourceSource": "Resource Source",
 		"Initiative": "Initiative",
 		"InitiativeCheck": "Initiative Check",
 		"InitiativeAbbr": "INIT",
--- a/module/checks/common-events.mjs
+++ b/module/checks/common-events.mjs
@@ -143,16 +143,19 @@ function calculateDamage(actor, item, config) {
  * @property {CharacterInfo} source
  * @property {UpdateResourceData} data
  * @property {FUItem} item
+ * @property {FUItemGroup} resourceSource
  * @property {CharacterInfo[]} targets
  * @property {CheckConfigurer} config
  */
 
 async function calculateResource(actor, item, config, data) {
+	const resourceSource = ItemUtils.resolveItemGroup(item);
 	const targets = config.getTargets();
 	const event = {
 		source: CharacterInfo.fromActor(actor),
 		targets: CharacterInfo.fromTargetData(targets),
 		item: item,
+		resourceSource: resourceSource,
 		data: data,
 		config: config,
 	};
--- a/module/checks/common-events.mjs
+++ b/module/checks/common-events.mjs
@@ -143,19 +143,19 @@ function calculateDamage(actor, item, config) {
  * @property {CharacterInfo} source
  * @property {UpdateResourceData} data
  * @property {FUItem} item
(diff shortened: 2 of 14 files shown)


module/checks/common-events.mjs
```diff
@@ -143,16 +143,19 @@ function calculateDamage(actor, item, config) {
  * @property {CharacterInfo} source
  * @property {UpdateResourceData} data
  * @property {FUItem} item
+ * @property {FUItemGroup} resourceSource
```
Azurelol: For this and other properties of this `FUItemGroup` type, let's rename them to just `itemGroup` so it's consistent across the different events and becomes clearer.

I forgot to do it!

module/documents/effects/triggers/calculate-resource-rule-trigger.mjs
```diff
@@ -37,6 +40,9 @@ export class CalculateResourceRuleTrigger extends RuleTriggerDataModel {
 				choices: Object.keys(FU.scalarChange),
 				required: true,
 			}),
+			resourceSources: new fields.SetField(new fields.StringField()),
```
Azurelol: Let's name this property `itemGroups` too. 

I will be renaming the `damageSources` on calculate damage rule trigger to match and do the migration.

module/expressions/expressions.mjs
```diff
@@ -405,6 +405,10 @@ function evaluateVariables(expression, context) {
 				}
 				return 0;
 			}
+			// High Roll of Check
+			case 'hr': {
+				return getHighRoll(context.check);
+			}
```
Azurelol, line 411: Nice! Tis exactly how I would have done it.

module/helpers/config.mjs
```diff
@@ -931,6 +931,7 @@ FU.itemGroup = {
 };
 
 FU.damageSource = FU.itemGroup;
+FU.resourceSource = FU.itemGroup;
```
Azurelol: For your new stuff use `FU.itemGroup`  directly; I will migrate the other RE to use `itemGroup` as well.

Azurelol (commented):
Great work. Just a few naming changes I would like you to do in order to align with an upcoming refactor of the calculate damage RE.

templates/effects/triggers/calculate-resource-rule-trigger.hbs
```diff
@@ -10,4 +10,21 @@
 	<select name="{{pfuConcat path ".change"}}">
 		{{selectOptions options.scalarChange selected=trigger.change sort=true localize=true}}
 	</select>
-</label>
\ No newline at end of file
+</label>
+
+<div class="fu-form__property">
+	<span>{{localize 'FU.ResourceSource'}}</span>
+	{{formGroup trigger.schema.fields.itemGroups value=trigger.itemGroups options=options.itemGroupOptions name=(pfuConcat path ".itemGroups") classes="fu-form__set stacked"}}
```
Azurelol: You will want to use the ItemGroup localization here too.

lang/en.json
```diff
@@ -91,6 +91,7 @@
 		"Resource": "Resource",
 		"ResourcePoints": "Resource Points",
 		"ResourceStatus": "Resource Points?",
+		"ResourceSource": "Resource Source",
```
Azurelol: You can remove this one, as all the item group properties across REs will be using the `ItemGroup` key.

Azurelol (approved):

PR #1076, opened 2026-04-20
Refactor damage source to item group

Azurelol:
The concept of resolving which "group" of items a particular item in an event belongs to is of particular interest to some of the key rule elements. When I first used it was in the damage events but it's become obvious to me this same information will be useful across other events and rule triggers, thus I am standardizing it here. (I meant to do this way back but like other things it fell by the wayside)

- Updates damage events to use `itemGroup` instead of `damageSource`.
- Migrates calculate damage source rule trigger's `damageSources` to `itemGroups`.
- Migrates damage rule triggers `damageSources` to `itemGroups`.

### DIFF
--- a/module/documents/effects/triggers/calculate-damage-rule-trigger.mjs
+++ b/module/documents/effects/triggers/calculate-damage-rule-trigger.mjs
@@ -7,7 +7,7 @@ const fields = foundry.data.fields;
 /**
  * @description Trigger based on a {@linkcode CalculateDamageEvent}
  * @extends RuleTriggerDataModel
- * @property {Set<FUItemGroup>} damageSources
+ * @property {Set<FUItemGroup>} itemGroups
  * @property {Set<DamageType>} damageTypes
  * @property {String} identifier
  * @property {Boolean} local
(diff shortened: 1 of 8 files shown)


module/documents/effects/triggers/calculate-damage-rule-trigger.mjs
```diff
@@ -28,14 +28,22 @@ export class CalculateDamageRuleTrigger extends RuleTriggerDataModel {
...
 		return schema;
 	}
 
+	static migrateData(source) {
+		if (source.damageSources) {
+			source.itemGroups = source.damageSources;
+			delete source.damageSources;
+		}
+		return super.migrateData(source);
+	}
```
Frozedon, line 45: Will this migrate the data for Rule Elements that are already on the PC's sheet?

Azurelol: Yes, any of them that are loaded from the DB will have this procedure invoked on each time until they are finally saved back. 

For this PR I would like you to verify on your end by doing a checkout and confirming that the migration is working for you too. I tested this with the `Cataclysm` spell which uses this.

Frozedon:
Tested with multiple Skills that have automation using the Calculate Damage and Calculate Expense Rule Triggers. Tested Volcano, Cheap Shot, Ceaseless Battlefield, and Follow My Lead. Both on an existing PC and from compendium. Looks all good to me. Not seeing any broken behavior.

Frozedon (approved):

PR #1077, opened 2026-04-20
Fix custom weapon trading and update the sheet

Azurelol:
- Update custom weapon table rendering on the stash/party sheets
- Fix custom weapon trading
- Update custom weapon sheet (to improve readability)

image

### DIFF
--- a/module/documents/items/skill/weapon-resolver.mjs
+++ b/module/documents/items/skill/weapon-resolver.mjs
@@ -224,4 +224,5 @@ export const WeaponResolver = Object.freeze({
 	getEquipment,
 	getEquippedWeapons,
 	getAccuracy,
+	normalizeData,
 });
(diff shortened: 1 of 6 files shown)


Frozedon:
There's no Accuracy or Damage displaying for Custom Weapons on the Party and Stash sheets.

image

image

Frozedon:
Also, when expanded, they don't display "tags" for Melee/Ranged and the weapon category.

image

Azurelol:
That was actually intentional since I didn't want to put in the extra work for handling transforming weapons at the time. But I guess I can just pick the first state if they are on that mode and call it a day.

Azurelol:
It's updated.

Frozedon (approved):
Code looks good to me. Also, tested in engine and not seeing any issues with the Custom Weapon display.

PR #1092, opened 2026-04-28
Fix for attribute select display issue with 2-digit dice values

Element-Re:
Somewhere in the layout changes I did for the actor sheet header, I introduced an issue wherein the attribute die select element doesn't show the 2nd digit in the case of 2-digit values. This PR resolves that issue.
Before:
image
After:
image

### DIFF
--- a/styles/css/actor/actor.css
+++ b/styles/css/actor/actor.css
@@ -624,6 +624,7 @@
 .projectfu .status-effect-toggle img {
 	height: 32px;
 	width: 32px;
+	object-fit: contain;
 	display: block;
 	margin-top: -1px;
 	border-radius: 4px;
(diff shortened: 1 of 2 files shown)


Unarekin (commented):
Hrm.  This does address the issue in the Electron client, and looks fine in Chrome (147.0.7727.55), it is a little squished in Firefox 150.0

Image

But honestly if I had to choose between it looking a little off for some subset of users vs. it not being readable for a different subset of users, I'd rather the former.  So if this isn't something you want to mess with at the moment I won't be too bothered with addressing that part of it some time later.

Element-Re:
I can probably tweak it a bit more so it looks better on all browsers. Fortunately I already spend some effort tweaking that sheet section to be a bit more flexible should the sizing of the elements within change.

Element-Re:
Additional changes to address the Firefox issue are in.

Unarekin (approved):

PR #1097, opened 2026-04-30
Disabled Status Immunity Setting When Overridden

### LINKED ISSUE #1019: Effect - Active while not in crisis
**Bug Description:**
The bug involves the predicate function of "not in crisis" when used to apply effect passively.

**Steps to Reproduce:**
1. Create an NPC with a passive that makes him immune to one status whilst not in crisis.
2. F5, Refresh Foundry.
3. Put that NPC on the scene.
4. Change his HP to be in crisis.
5. The NPC stays immune to the effect.

"Hotfix"
3. Before putting the NPC on the scene, go to the effect page, and click on the checkbox for the immunity to remove it. The passive puts it back immediately.
4. Put the NPC on the scene.
5. Put his HP above and below threshold, notice it works just fine, it turns on and off.

From there, if you refresh, the bug will resume, and the immunity will stay active even in crisis.

**Expected behavior:**
I suppose I would expect the passive to properly turn on and off when crisis is reached or left.

I use the same line as the one provided in actors effects for immunities;

Image

Frozedon:
Added a conditional to the Status Immunity checkboxes of the actor settings. It disables the checkbox if the Status Immunity is currently being overridden by an active effect. That way the state of the Status Immunity is not being saved to the actor while it's being controlled by an active effect.

This fixes #1019

### DIFF
--- a/templates/actor/character/parts/actor-section-settings.hbs
+++ b/templates/actor/character/parts/actor-section-settings.hbs
@@ -245,7 +245,10 @@
 								<span class="resource-label-sm">{{immunity.label}}</span>
 								<input type="checkbox"
 									   id="immunity-{{key}}"
-									   name="system.immunities.{{key}}.base"
+									   name="system.immunities.{{key}}.base"  
+									{{#if (lookup @root.actor.overrides.system.immunities key)}}
+									   disabled data-tooltip="{{localize "FU.DisabledByActiveEffect"}}"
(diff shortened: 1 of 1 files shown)


Unarekin (approved):
Approving, though we may consider shifting fields in the settings page to use `_source` in the future in order to bypass ActiveEffect changes

PR #1104, opened 2026-05-02
Pressure system affinities

### LINKED ISSUE #1096: Resistance Not Being Used Correctly With Pressure Clock
**Bug Description:**
When an NPC with a Pressure Clock suffers damage they are Resistant to and they have a Pressure Point for, the Resistance is ignored. Also, the amount to be filled is checking damage instead of HP Loss.

**Steps to Reproduce:**
1. Create a PC.
2. Give that PC a Staff.
3. Create a NPC.
4. Set the NPC's Affinity to have Resistance to Physical damage.
5. Set the NPC's Pressure Point for Arcane.
6. Make the NPC a Champion or Elite.
7. Start an Encounter with both the PC and NPC.
8. Have the PC perform an attack with the Staff targeting the NPC.
9. Apply Damage to the NPC.
10. Observe that Resistance was ignored.**Expected behavior:**
Resistance and other Affinities should be taken into account when damage is applied, even if dealt with a Pressure Point.

Image

Unarekin:
Per the Pressure & Stagger optional rule, Pressure Points are evaluated on HP loss rather than damage, which means they are handled *after* Affinities are applied.  Vulnerability is excepted from this, as their Pressure clock is increased *instead* of doubling HP loss.

This PR adjusts the conditions where affinities are applied during the damage pipeline to allow for applying Resistance, as well as bypassing Pressure when an NPC is Immune or Absorbs a damage type, as those would not trigger HP loss.

image

### DIFF
--- a/module/pipelines/damage-pipeline.mjs
+++ b/module/pipelines/damage-pipeline.mjs
@@ -245,22 +245,26 @@ function resolveAffinity(context) {
 		if (game.settings.get(SYSTEM, SETTINGS.pressureSystem) && context.actor.resolveEffect('pressure')) {
 			let pressurePointTriggered = false;
 
-			// Check if the damage type hit a vulnerable affinity
-			const vulnerabilityTriggered = context.affinity === FU.affValue.vulnerability;
-			if (vulnerabilityTriggered) {
-				context.pressureTrigger = StringUtils.localize(FU.damageTypes[context.damageType]);
-				pressurePointTriggered = true;
-			}
-			// Do a lookup on the weapon used, whose traits are passed on to the context.
-			else {
-				/** @type NpcDataModel **/
-				const npcData = context.actor.system;
-				for (const trait of context.traits) {
-					if (npcData.pressurePoints.has(trait)) {
-						pressurePointTriggered = true;
-						context.pressureTrigger = TraitUtils.localize(trait);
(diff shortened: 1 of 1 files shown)


module/pipelines/damage-pipeline.mjs
```diff
@@ -245,22 +245,26 @@ function resolveAffinity(context) {
...
-				const npcData = context.actor.system;
-				for (const trait of context.traits) {
-					if (npcData.pressurePoints.has(trait)) {
-						pressurePointTriggered = true;
-						context.pressureTrigger = TraitUtils.localize(trait);
-						context.pressurePoint = trait;
-						break;
+			// Since pressure points are triggered on HP loss, simply bypass this entire check when
+			// Immune or Absorb, since HP is not lost in these cases.
+			if (context.affinity !== FU.affValue.immunity && context.affinity !== FU.affValue.absorption) {
```
Frozedon, line 250: Does this check take into account if the source of damage is ignoring Immunities or Absoprtion? Like with the "IgnoreImmunities", "IgnoreAbsoprtion", or "IgnoreAffinities" traits.

Unarekin: It does.  The Ignore- traits are handled just above this in the pipeline

Frozedon (approved):

PR #1110, opened 2026-05-04
Compendium Browser Sorting

### LINKED ISSUE #781: Feature: Sorting Options for Compendium Browser
**User Story:**
As a player, I want to be able to choose the sorting order for the items displayed in the Compendium Browser.

**Additional context:**
Currently, all the items in the Compendium Browser are in a haphazard, somewhat random order. By default, we should sort alphabetically by name. Though we could also have options for sorting alphabetically by "Class" or by "Source". Also, for the Adversaries tab, sorting by "Species", "Role", or "Rank".

**Attachments:**

Image

Frozedon:
Applied the sorting function for the Compendium table into the correct location. Also, added the correct classes to the compendium browsers so they have the scrolling capability.

This partially fixes #781 to at least have a default of alphabetical.

image

### DIFF
--- a/module/ui/compendium/compendium-browser.mjs
+++ b/module/ui/compendium/compendium-browser.mjs
@@ -20,7 +20,7 @@ class CompendiumTableRenderer extends FUTableRenderer {
 	static TABLE_CONFIG = {
 		getItems: async (entries) => entries.filter((e) => e.name),
 		tablePreset: 'compendium-item',
-		sort: (a, b) => a.system.fuid.localeCompare(b.system.fuid),
+		sort: (a, b) => a.name.localeCompare(b.name),
 	};
 }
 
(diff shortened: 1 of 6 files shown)


module/ui/compendium/compendium-browser.mjs
```diff
@@ -20,7 +20,7 @@ class CompendiumTableRenderer extends FUTableRenderer {
 	static TABLE_CONFIG = {
 		getItems: async (entries) => entries.filter((e) => e.name),
 		tablePreset: 'compendium-item',
-		sort: true,
+		sort: (a, b) => a.system.fuid.localeCompare(b.system.fuid),
```
Unarekin: Better here to sort by the actual name rather than fuid

While sorting by fuid will usually get the same results in English, this is not necessarily the case and is most assuredly not the case in other languages.

Image

Unarekin (changes requested):

module/ui/compendium/compendium-browser.mjs
Frozedon: Updated
image

Unarekin (approved):

Issue #1122, opened 2026-05-15
"Check" rule predicate triggers when it should not


Shourn:
**Bug Description:**
The "Check" rule predicate triggers when it should not, most easily seen with the "Loremaster" skill "Flash of Insight".

**Steps to Reproduce:**
1. Get 'Flash of Insight' from the compendium
2. Roll an Open Check using Insight for one but not both attributes
3. Observe 'Flash of Insight' message in the chat

**Expected behavior:**
The 'Flash of Insight' message should only trigger for Open Checks that specifically use [INS+INS].
It also should only trigger for rolls of 13 or higher.

**Attachments (if applicable):**
<img width="302" height="399" alt="Image" src="https://github.com/user-attachments/assets/e0b9c284-2099-499d-99c4-32d5617978b9" />

Frozedon:
Compendium of Flash of Insight Skill is fixed. The next step will be to clean up the Check Rule Predicate with #1190

PR #1123, opened 2026-05-15
bugfix: "Check" rule predicate triggers when it should not

Shourn:
- temporarily remove "result" functionality from CheckRulePredicate, it does not work anyway
- improve 'check attributes' predicate functionality to properly account for duplicate and single attributes
- adjust 'Flash of Insight' notification wording

Related to #1122, but does not fully fix the problem

### DIFF
--- a/module/documents/effects/predicates/check-rule-predicate.mjs
+++ b/module/documents/effects/predicates/check-rule-predicate.mjs
@@ -159,14 +159,14 @@ export class CheckRulePredicate extends RulePredicateDataModel {
 		}
 
 		// Check attributes
-		const a = [this.attributes.primary.value, this.attributes.secondary.value];
-		const b = context.eventType === FUHooks.PERFORM_CHECK_EVENT ? [check.primary, check.secondary] : [check.primary.attribute, check.secondary.attribute];
-		for (let i = 0; i < a.length; i++) {
-			const attribute = a[i];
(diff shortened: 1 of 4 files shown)


Frozedon:
Does the result part work for "Render Check" Rule Triggers? It may not be there for "Perform Check", but i believe the result should be available on Render Check.

Frozedon:
Doing a quick look. It looks as though Flash of Insight should be using the "Render Check" Rule Trigger, not the "Perform Check" rule trigger. It is concerning that "Perform Check" is allowing the "Check" Rule Predicate to be applied, since as you've pointed out, it doesn't have a result to actually check.

module/documents/effects/predicates/check-rule-predicate.mjs
```diff
@@ -159,18 +159,25 @@ export class CheckRulePredicate extends RulePredicateDataModel {
...
 
-		// Check result
-		if (this.result != null && check.result <= this.result) {
-			return false;
-		}
+		// FIXME: Rule element workflow cannot access check results because the check has not yet been rolled when the predicate is checked
+		// // Check result
+		// if (this.result != null && check.result <= this.result) {
+		// 	return false;
+		// }
```
Frozedon: I don't believe this is the fix because "Render Check" Rule Trigger does have the result available to check against. It's only "Perform Check" thats problematic.

Frozedon (changes requested):

Frozedon (approved):

PR #1125, opened 2026-05-16
Item transfer via Drag&Drop

Shourn:
It's a bit of a niche situation but I had several instances where players are owners of both their character and the party actor and trying to transfer items via drag&drop caused the item to be duplicated. With this change the drag&drop workflow causes the item to get traded through the inventory pipeline as usual.


Frozedon (approved):
I've also ran into this issue. I think this is a good change because users seem to expect it to "move" the item, not just copy it

PR #1143, opened 2026-05-19
Fix Theriomorphosis and Arcanum Applications

Frozedon:
- CommonEvents.feature needs an empty array to fill the targets parameter.
- Fixed expenses to not strip them when going through the ResourcePipeline.calculateExpense function.


Frozedon:
Already approved in main, just ferrying to dev as well

PR #1146, opened 2026-05-19
feat: Support special actions and rules in NPC profile

### LINKED ISSUE #452: Adversary Tab to allows viewing of NPC's special rules and actions
**User Story:**
As a GM], I want check/uncheck which special rules/special actions the players can see to carter for Flash of Insights or when the player encounters them.

**User Examples:**
As a GM, I want checkboxes for each of the special rule/actions or notes for a NPC so that I can reveal them to the PCs progressively, perhaps due to Flash of Insights, or an project. For example, some of the bosses in the various atlas has lots of special rules and actions and as of now PCs must track them manually.

### LINKED ISSUE #929: Adding Other Action in Study Information
Simple thing. When you study an enemy it shows Basic Attack, Weapon and Spell.

Could it show the full equipment, other action, rules listing too? For exemple I use Other Action for one of my humanoid enemy and a lots of my special abilities for enemies also use it. When a player study it, all he know its that the enemy has a weapon but can't learn about their attack and rules.

Image

Shourn:
- allow GMs to reveal abilities, rules and status immunties in NPC profiles

resolves #452
resolves #929

### DIFF
--- a/module/ui/npc-profile.mjs
+++ b/module/ui/npc-profile.mjs
@@ -166,7 +166,7 @@ export class NpcProfileWindow extends FUApplication {
 		context.statusEffects = FU.statusEffects;
 		context.statusImmunities = Object.entries(actor.system.immunities)
 			.filter(([, immune]) => immune)
-			.filter(([status]) => data.revealed?.statusImmunities[status])
+			.filter(([status]) => data.revealed?.statusImmunities?.[status])
 			.map(([status]) => status);
 
 		context.basicAttacksTable = await this.#basicAttacksTable.renderTable(actor);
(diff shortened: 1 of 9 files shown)


module/ui/npc-profile.mjs
```diff
@@ -146,13 +162,18 @@ export class NpcProfileWindow extends FUApplication {
 		context.level = system.level.value;
 		context.hp = system.resources.hp.max;
 		context.mp = system.resources.mp.max;
-		context.localizedSpecies = FU.species[this.data.species];
+		context.localizedSpecies = FU.species[data.species];
+		context.statusEffects = FU.statusEffects;
+		context.statusImmunities = Object.entries(actor.system.immunities)
+			.filter(([, immune]) => immune)
+			.filter(([status]) => data.revealed?.statusImmunities[status])
```
Unarekin: Probably ought to have an optional chaining operator on `statusImmunities` here.

Trying to open the profile of adversaries without revealed status immunities throws an error
- If the adversary is from prior to this change and hasn't been updated after
- No data is revealed at all
- Data has been revealed after this change, but it has no status immunities in the first place

Unarekin (changes requested):

module/ui/npc-profile.mjs
Shourn: Good catch. I must've tunnel visioned on the happy path.

Unarekin (approved):

Unarekin (approved):

PR #1160, opened 2026-05-21
feat: Confirmation dialog for 'weaponMagicCheck' functionality

### LINKED ISSUE #1002: Make Spellblade's effect Optional on spellcast
**User Story:**
As a player, I want Spellblade to be optional when casting a spell so that I can choose between INS+WLP or Weapon Accuracy on spellcast.

**User Examples:**
The Text for Spellblade uses "May" meaning the user may choose to use it's effect or not when casting a spell. In practical terms if suffering from Debuffs that hit weapon accuracy while benefiting from buffs to INS or WLP, even a character built to use Spellblade may prefer to use their default casting stats sometimes.

**Additional context:**
I appreciate that this adds some friction to casting spells by adding an extra click, and most Spellblades will be built to use Spellblade 95% of the time. But considering Cataclysm already will add an extra decision on every spellcast, I don't think it would be too onerous an addition, especially if it can share the same dialogue

Although if it could hijack something like the damage customizer where a modifier key can be held when activating a spell to make these more niche decisions, that would be even better.

Shourn:
- added confirmation prompt for using weapon accuracy for a magic check when 'weaponMagicCheck' flag is set

fixes #1002 

It's not the most elegant solution and can't accurately model the rules yet, but it works for now.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1999,6 +1999,8 @@
 		"DialogCreateItemSelectTypeContent": "Select the type of item you want to create:",
 		"DialogCreateItemSelectTypeTitle": "Select Item Type",
 		"CompendiumMigrateItemKeepImages": "Keep item images unchanged",
-		"CompendiumMigrateItemKeepNames": "Keep item names unchanged"
+		"CompendiumMigrateItemKeepNames": "Keep item names unchanged",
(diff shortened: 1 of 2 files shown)


Unarekin (approved):
Should be good to go, provided the merge conflict in `lang/en.json` is addressed.

PR #1169, opened 2026-05-23
Adversary list search and filter

Unarekin:
Adds search and filter options to the adversary list on the party sheet

image

### DIFF
--- a/module/sheets/actor-party-sheet.mjs
+++ b/module/sheets/actor-party-sheet.mjs
@@ -28,6 +28,7 @@ import { ProgressPipeline } from '../pipelines/progress-pipeline.mjs';
 import { FUBondChart } from '../ui/bond-chart.mjs';
 import { CodexBrowser } from '../ui/codex-browser.mjs';
 import { VehicleModuleTableRenderer } from '../helpers/tables/vehicle-module-table-renderer.mjs';
+import { HTMLUtils } from '../helpers/html-utils.mjs';
 
 /**
  * @description Creates a sheet that contains the details of a party composed of {@linkcode FUActor}
@@ -151,6 +152,9 @@ export class FUPartySheet extends FUActorSheet {
 	#codexDrop;
 	/** @type SheetExtensions **/
 	#extensions;
+	#lastAdversarySearch = '';
+
+	#adversaryFilters = {};
 
 	constructor(...args) {
 		super(...args);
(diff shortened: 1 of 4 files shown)


module/sheets/actor-party-sheet.mjs
```diff
@@ -213,6 +217,19 @@ export class FUPartySheet extends FUActorSheet {
 			xp: experience.total,
 			zenit: this.party.resources.zenit.value,
 		};
+
+		// Ranks are purposefully not sorted, so they appear in natural order
+		context.adversaryRanks = Object.entries(FU.rank).map(([key, value]) => ({ value: key, label: value }));
+		context.adversarySpecies = Object.entries(FU.species)
+			.sort((a, b) => a[1].localeCompare(b[1]))
+			.map(([key, value]) => ({ value: key, label: value }));
```
Shourn: I propose sorting by localized name instead of translation key.
```suggestion
		context.adversarySpecies = Object.entries(FU.species)
			.map(([key, value]) => ({ value: key, label: game.i18n.localize(value) }))
			.sort((a, b) => a.label.localeCompare(b.label));
```

module/sheets/actor-party-sheet.mjs
Shourn, line 1: With the current implementation only the search field state is restored after a render.
I feel like the state of the checkbox filters should also be restored, especially since changes to PC actors will cause the party sheet to render without warning.

module/sheets/actor-party-sheet.mjs
```diff
@@ -354,6 +415,37 @@ export class FUPartySheet extends FUActorSheet {
...
 		switch (partId) {
+			case 'adversaries':
+				{
+					const searchInput = this.element.querySelector('#adversarySearch');
+					if (searchInput instanceof HTMLInputElement) {
+						searchInput.addEventListener(
+							'input',
+							HTMLUtils.debounce(() => {
+								const text = searchInput.value.toLowerCase() || '';
+								this.#lastAdversarySearch = text;
```
Shourn: Since `this.#lastAdversarySearch` is used to restore the search field state after a render I would refrain from changing the user supplied value.
I am all in favor of the search being case-insensitive, but changing the user input in the process may cause confusion.

module/sheets/actor-party-sheet.mjs
```diff
@@ -344,6 +361,50 @@ export class FUPartySheet extends FUActorSheet {
...
+			if (this.#adversaryFilters.misc) {
+				for (const misc of this.#adversaryFilters.misc) {
+					switch (misc) {
+						case 'inCombat':
+							if (!game.combat.hasInstancedActor(adversary.uuid)) display = false;
+					}
+				}
+			}
+
+			element.style.display = display ? 'flex' : 'none';
```
Shourn: I would prefer using the core-foundry utility CSS class `hidden` instead of mutating the style property.
```suggestion
			element.classList.toggle("hidden", !display);
```

Shourn (changes requested):
I won't give you too much trouble because of the inline styles, flexboxes and overflow are a notoriously tricky combination.
Nonetheless, I would appreciate if you could get rid of them.

Shourn (approved):

PR #1175, opened 2026-05-26
Compendium browser spell filters

Unarekin:
This PR updates the filter list on the Spells tab of the Compendium Browser to use class fuids for checkbox values, rather than names.
It also does a lookup for the classes to ensure it retrieves their proper localized name
And it localizes the NPC list item

We may need to consider doing something like adding a boolean field to the class data model to indicate if that class grants spells or not, so that this filter list can be generated automatically instead of using a hard-coded set of values

### DIFF
--- a/module/ui/compendium/compendium-browser.mjs
+++ b/module/ui/compendium/compendium-browser.mjs
@@ -800,14 +800,22 @@ export class CompendiumBrowser extends FUApplication {
 
 			case 'spells':
 				{
-					const spells = await this.index.getItemsOfType('spell');
-					const classes = ['Elementalist', 'Entropist', 'Spiritist', 'NPC']; // hardcoded for now
-					const classOptions = classes.map((c) => {
-						return {
-							value: c,
-							label: c,
-						};
-					});
+					const [spells, classes] = await Promise.all([this.index.getItemsOfType('spell'), this.index.getItemsOfType('class')]);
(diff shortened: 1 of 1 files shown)


module/ui/compendium/compendium-browser.mjs
```diff
@@ -800,14 +800,22 @@ export class CompendiumBrowser extends FUApplication {
...
+							return {
+								value: c,
+								label: classes.find((item) => item.system.fuid === c)?.name ?? c,
+							};
+						}),
+						{
+							value: 'NPC',
+							label: 'ACTOR.TypeNpc',
+						},
+					];
```
Shourn: I admit the solution for NPC Spells is similarly brittle to the old way, but at least this would handle homebrew spellcasters without problems.
```suggestion
					const classesWithSpells = spells
						.map((spell) => spell.system.class.value)
						.reduce((acc, cur) => {
							acc.add(cur);
							return acc;
						}, new Set());
					const classOptions = [
						...classesWithSpells.map((c) => {
							let label;
							if (c.toLowerCase() === 'npc') {
								label = 'ACTOR.TypeNpc';
							} else {
								label = classes.find((item) => item.system.fuid === c)?.name ?? c;
							}
							return {
								value: c,
								label: label,
							};
						}),
					];
```

Shourn (changes requested):

Shourn (approved):

PR #1178, opened 2026-05-26
fix: run class fuid converter on item creation

rsbarroco:
## Summary

`ClassFuidConverter` currently runs only once during the system `ready` hook. Skills, spells, or heroic items added to actors mid-session keep their original `system.class.value` (which can be the class display name with capitalization, e.g. `"Predator"`) and never get normalized to the matching class `fuid` (`"predator"`). The character sheet's class level tally compares the slugified class name against this stored value, so the affected class silently reports `0/10` on the Classes tab even though the player has invested skill ranks in it.

This PR makes two narrow adjustments:

1. **Run the converter on `createItem`.** A `createItem` hook in `projectfu.mjs` calls a new `ClassFuidConverter.convertItem(item)` helper so that newly added skills/spells/heroic items are normalized immediately instead of having to wait for the next world reload.
2. **Don't set `flags.projectfu.classConverted` on unresolved items.** Previously the flag was set unconditionally in `convertClassAttributeFromClassLists`, even when the converter could not find a matching class (e.g. compendium not yet loaded, class not yet added to the world). That marked the item as migrated and prevented retries on subsequent sessions. The flag is now only set when every class reference resolved.

## Repro

1. Create a character with class A (whose `fuid` matches its display name slugified).
2. Add class B to the same character (where class B is a class whose skills ship with `system.class.value` using a different casing than the class `fuid`, e.g. `Predator` skills from compendium content).
3. Add one of class B's skills to the character.
4. Open the Classes tab: class B shows `0/10` although the skill has ranks.

After the fix, the new skill's `system.class.value` is normalized at creation time and the tally is correct without reloading the world.

## Notes

- The conversion logic itself is unchanged; only its invocation pattern.
- The pre-existing `ClassFuidConverter.run()` migration sweep on `ready` still runs as before, so worlds opened before this change still get caught up on the next launch (and now correctly retry items that previously failed silently).
- No new dependencies, no schema changes, no breaking changes to the public API of `ClassFuidConverter` (`run()` kept; `convertItem()` added).

### DIFF
--- a/module/documents/items/class-fuid-converter.mjs
+++ b/module/documents/items/class-fuid-converter.mjs
@@ -3,13 +3,8 @@ import { Flags } from '../../helpers/flags.mjs';
 import { CompendiumIndex } from '../../ui/compendium/compendium-index.mjs';
 
 async function convertClassAttributeFromClassLists(item, classes) {
-	let updates = {
-		flags: {
-			[SYSTEM]: {
-				[Flags.ClassConverted]: true,
-			},
-		},
-	};
+	let updates = {};
+	let conversionResolved = true;
 
 	const classAttribute = item.system?.class?.value;
 	if (classAttribute) {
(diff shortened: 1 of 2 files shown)


Shourn:
After thorough discussion we have come to the conclusion, that this is not something that needs to be fixed in the system.
Rather, this is a problem with the compendiums the data originated from.
We have already started updating the homebrew module, which we assume is where you got the class and skills from, and should have an updated version within the next few days.

PR #1200, opened 2026-06-05
Feature: Align rules to miscAbilities in terms of configurability

### LINKED ISSUE #1078: Basic Automation Capability for Special Rules
**User Story:**
As a gm, I want it to be easy to create basic automation for Special Rules. To have them make Checks, apply Damage, gain/lose Resources, enact a Cost, and apply Effects. All that would display and have enacting buttons for when the Special Rule is performed and sent to chat.

**Additional context:**
Special Rules on NPCs can do about anything. It would be good to support a series of basic functionality directly though the Special Rule's Attributes. We already have these method available in PC Skills, and it would be best to replicate all that functionality to NPC Special Rules.

The functions we should support:
- **Making a Check**. Setting the Attributes, adding a bonus, determining the Defense type to target.
- **Cost**. Set the Resource type, Amount, and if it should scale with number of Targets.
- **Damage**. Damage type, bonus to damage, if it should include HR or not.
- **Resource**. Resource Type and Amount.
- **Effects**. What Effects can this apply, any adjustments to those effects, and if it should prompt for choice of effects.
- **Targeting**. What targeting Rule to use.
- **Traits**. Any Traits the Special Rule should apply with.

**Attachments:**

Personally would prefer if it looked mostly like the existing PC Skill Attributes.

Image

Shourn:
- factored out "skill-like" item handling
- added more sections to rules
- adjusted rule table rendering

fixes #1078


Frozedon (changes requested):
Issues
- Adding Special Rule  on NPC sheet "+Add" button adds Misc Ability instead.
- Misc Ability doesn't display anything other than title and tags when performed.
- Special Rules should be considered as part of the Skill ItemGroup for the purposes of Rule Elements.
- Special Rule should not have the Skill action Gear displaying on the sheet preview.
- Clicking the Ping Target button on a Special Rule errors in the console.

Frozedon (approved):

PR #1212, opened 2026-06-07
Bugfix: Spellblade not properly accounting for skill level

### LINKED ISSUE #1199: Spellblade doesn't account for weapon's bonus accuracy
**Bug Description:**
Spellblade doesn't apply accuracy bonuses for the equipped weapon to its own accuracy formula, such as the +2 bonus when using a custom weapon with the Accurate customization.

Shourn:
- fix Spellblade check bonus required attributes incorrectly initializing to [DEX + DEX]

I was unable to reproduce the issue in #1199 but found that Spellblade does not correctly add the skill level as bonus to check formulas that include dexterity only once and fixed that.

fixes #1199

### DIFF
--- a/module/documents/effects/predicates/check-rule-predicate.mjs
+++ b/module/documents/effects/predicates/check-rule-predicate.mjs
@@ -1,7 +1,6 @@
 import { RulePredicateDataModel } from './rule-predicate-data-model.mjs';
 import { systemTemplatePath } from '../../../helpers/system-utils.mjs';
 import { FU } from '../../../helpers/config.mjs';
-import { ItemAttributesDataModel } from '../../items/common/item-attributes-data-model.mjs';
 import { FUHooks } from '../../../hooks.mjs';
 import { CheckConfiguration } from '../../../checks/check-configuration.mjs';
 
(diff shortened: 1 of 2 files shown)


NateMcCloud:
The way I did it was putting a number in the equipped custom weapon's Bonus field in the Accuracy formula), then using a spell and choosing to use the weapon's accuracy formula. Normally, the +2 is added in the Mod field, but it doesn't work that way when casting a spell.

Will have to test it next time the GM updates his PFU version.

Shourn:
Please do test and get back to us.
My testing with both normal and custom weapons always had their respective accuracy bonuses applied as expected.

NateMcCloud:
> Please do test and get back to us. My testing with both normal and custom weapons always had their respective accuracy bonuses applied as expected.

Testing it on my own machine, it does work as intended. I think the GM in the game I was testing it on before may have been using an older version or something.

Frozedon (approved):

Frozedon (changes requested):
Needs bigger fix

Frozedon (approved):

PR #1217, opened 2026-06-10
Automatically open support check prompt

### LINKED ISSUE #1201: Faster Group Checks
As a player, I want Group Checks to be faster so that more time can be spent on roleplay.

When an actor initiates a Group Check, it should automatically prompt the same check for all other actors listed on the party sheet. This would make the process quicker and reduce the number of manual steps required.

This functionality should not replace the current chat-card workflow. The Group Check chat card should remain available so players can open or reopen it if it was closed accidentally.

Ideally, this feature would be optional on a per-player basis, allowing each user to decide whether these prompts automatically appear on their screen. As such, it should remain integrated with the existing chat-card system.

If possible, the same functionality could also be applied to Initiative Checks. In that case, the prompts would be sent to actors listed in the Combat Tracker rather than those on the party sheet.

Thanks a lot for all the work already in the system!

Shourn:
- add setting controlling whether the user gets automatically prompted for a Support Check whenever another user starts a Group Check
- for Initiative Checks only prompt players who take part in combat

fixes #1201

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -2014,6 +2014,8 @@
 			"ChangeWarning1": "Proceed with caution!"
 		},
 		"NpcEquipmentSettings": "NPC Equipment",
-		"NpcEquipmentSettingsHint": "Controls whether NPCs can acquire the 'Use Equipment' skill, and the initial skill points available for 'Humanoid' NPCs."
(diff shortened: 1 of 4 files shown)


Frozedon (commented):
Can we make two different setting one for general Groups Checks and the other for Initiative Checks? I could see wanting to automatically having everyone support initiative, but not wanting the same for Group Checks since its not assumed everyone is going to support a narrative group check.

Frozedon (approved):

PR #1225, opened 2026-06-13
Modify Damage Action Compatible With Damage Event

### LINKED ISSUE #983: Feature: Rule Action to Modify Damage on Application
**User Story:**
As a dev, I want to have a Rule Action that can be used for the "Damage" Rule Trigger that is able to modify the damage being applied.

**User Examples:**
As a dev, I want to automate the "Hunter" rare weapon quality. It allows the weapon to deal 5 extra damage to creatures of a particular species. Since this is specific to each individual character being damaged, it can only be accurately evaluated on damage application.

Image

As a dev, I want to automate the extra damage for the Clatterclown's Jester Kick (Core p328). It deals 5 extra damage against dazed or shaken targets. 

Image

### LINKED ISSUE #1013: Rule action - Damage change
**User Story:**
As a gm, I would like to have a "Modify Damage" Rule action to accommodate for the automation of certain skills and rules.

**User Examples:**
As a gm, I would like to be able to make a rule being able to automatically adjust damage dealt.
For example, make it so that a creature would take no damage when they have no status effect, take X more damage when under a special effect, or take less damage when the attacking player his under certains effect.

Frozedon:
Updated the Damage Pipeline and the "Modify Damage" Rule Action to allow Rule Elements to modify damage as its being applied to a creature. That way extra damage based on conditions (such as species or status effect) can be applied to specific creatures.

Removed the "effect" option from the "Modify Damage" Rule Action since it didn't make sense for that option to be there. Also, nothing in the compendium used it, except for the "Venom" Effect in Infusions that didn't work anyway. Removed all the non-working effects from the Infusions Class Feature.

Updated with the "Check" Rule Predicate to check for both possible data structures of the target list.


module/checks/common-events.mjs
```diff
@@ -74,40 +75,33 @@ function attack(inspector, actor, item) {
...
- * @property {Number} amount
- * @property {CharacterInfo} target
- * @property {FUItem} item
- * @property {FUActor} actor
- * @property {Token} token
- * @property {Set<String>} traits
- * @property {FURenderData} renderData
+ * @property {FUActor} targetActor
+ * @property {DamagePipelineContext} damageContext
+ * @property {DamageType} damageType
```
Shourn: This documentation change appears to be wrong: This is the type definition for `DamageEvent`, not the documentation for the `damage` function.
It looks like the only change to `DamageEvent` is the addition of `damageContext`.

module/documents/effects/predicates/check-rule-predicate.mjs
```diff
@@ -120,8 +120,8 @@ export class CheckRulePredicate extends RulePredicateDataModel {
 									case 'all':
 										if (
 											!selected.every((target) => {
-												if (target.data.result === 'hit') return this.outcome === 'success';
-												if (target.data.result === 'miss') return this.outcome === 'failure';
+												if (target.data?.result === 'hit' || target.result === 'hit') return this.outcome === 'success';
```
Shourn: Not a huge deal but having the same condition repeated 6 times makes me want to factor it out.
Something like this:
```javascript
/**
 * @param {"hit", "miss"} result
 * @param {object} target
 */
function isResult(result, target) {
  return target.data?.result === result || target.result === result;
}
```

Shourn (changes requested):

Shourn (approved):

PR #1228, opened 2026-06-13
Bugfix: Unable to distinguish between temporary effects applying to Actor and Item

### LINKED ISSUE #1227: Unable to distinguish temporary effects that apply to an actor from those that do not.
**Bug Description:**
If I have an item that has a toggleable "equipped" state and add a temporary effect to it, that effect will always be displayed in the actor sheet "Temporary Effects" table and it is not possible to distinguish if it applies to the actor or not.

**Steps to Reproduce:**
1. Create Actor
2. Add a new Weapon to Actor
3. Add temporary Active Effect to Weapon
4. Observe the Active Effect being shown in the Actor sheet "Temporary Effects" as if it was applied to the Actor

**Expected behavior:**
It should be possible to clearly distinguish between the following three categories of temporary effects:
1. temporary Active Effects that apply to an Item instead of the Actor but are shown for informational value (most notably temporary weapon enchantments)
2. temporary Active Effects that apply to the Actor
3. temporary Active Effects that can apply to the Actor but currently are not applied because they are not transferred from the Item

Shourn:
- add new table for temporary effects on items to actor sheet
- fix suppression styling not being applied

fixes #1227

### DIFF
--- a/templates/common/active-effects.hbs
+++ b/templates/common/active-effects.hbs
@@ -12,6 +12,8 @@
 		</header>
 	{{/unless}}
 
+	{{{itemTemporaryEffectsTable}}}
+
 	{{{temporaryEffectsTable}}}
 
 	{{{passiveEffectsTable}}}
--- a/templates/common/active-effects.hbs
+++ b/templates/common/active-effects.hbs
@@ -12,9 +12,9 @@
 		</header>
 	{{/unless}}
 
-	{{{itemTemporaryEffectsTable}}}
-
 	{{{temporaryEffectsTable}}}
+
+	{{{itemTemporaryEffectsTable}}}
 
 	{{{passiveEffectsTable}}}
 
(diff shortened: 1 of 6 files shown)


templates/common/active-effects.hbs
```diff
@@ -12,6 +12,8 @@
 		</header>
 	{{/unless}}
 
+	{{{itemTemporaryEffectsTable}}}
+
 	{{{temporaryEffectsTable}}}
```
Frozedon: Could we swap these?

Frozedon (changes requested):
I think it would be better if "Temporary Effects" were above "Temporary Effects on Items". Mainly because most of the Temporary Effects are going to be active on the character while the On Items will be all inactive.

Image

Frozedon (approved):

PR #1229, opened 2026-06-16
Condense Checks on Chat Card + Updated Reroll/Push Windows

Frozedon:
- Made Accuracy, Magic, and general Check's Attributes more compact. They now don't take up as much vertical space.
- Fixed margins on descriptions and extra messages so they don't have as much wasted space in description box.
- Updated hover text on Accuracy/Magic Checks.
- Reroll window now displays each Attribute's die size and clearly labels the result.
- Added "Ignore Fabula Point Cost" option to Push window.
- Added DL to Target panel to compare to for Hit and Miss.
- Fixed the centering on the "Ping Target" icon.

## Screenshots

image
image
image
image
image
image


module/checks/check-configuration.mjs
```diff
@@ -549,7 +549,7 @@ export class CheckConfigurer extends CheckInspector {
 					targetResult = 'hit';
 				} else if (this.check.fumble) {
 					targetResult = 'miss';
-				} else {
+				} else if (difficulty) {
```
Shourn: I'd prefer if the whole for-loop was moved into an `if (targetedDefense) {}` block instead of this change.
If there is no targeted defense we don't need to evaluate this whole block, but if a defense is targeted it generally _should_ be targetable and result in a hit/miss decision.

Unless I'm missing an edge case here?

module/checks/check-reroll.mjs
```diff
@@ -141,6 +143,10 @@ const getRerollParams = async (check, actor) => {
 
 				const ignoreFp = dialog.element.querySelector('input[name="ignore-fp"]').checked;
 
+				if (!trait) {
+					return false;
+				}
+
```
Shourn: I think it would be better to use `trait: trait?.value` and `value: trait?.dataset.value ?? ''` in the return, than short circuiting here.
If we return `false` from the dialog, none of the user feedback notifications in lines 161-170 get triggered, most notably the "you forgot to select a trait" notification.

styles/css/utils/icons.css
```diff
@@ -558,6 +558,11 @@
 	--icon-fu-mig: url('../../static/icons/attributes/mig-glyph.png');
 	--icon-fu-wlp: url('../../static/icons/attributes/wlp-glyph.png');
 
+	--icon-fu-dex-bl: url('../../static/icons/attributes/dex-borderless.png');
```
Shourn, line 585: Please also use the `@property` syntax to define these, it enables better usability of URL-type properties.
See above for how to do it.
Also note that the "double initialization" is deliberate and needed because Chromium and Firefox have slightly different behavior for `@property` default values.

templates/chat/partials/chat-accuracy-check.hbs
```diff
@@ -20,16 +20,16 @@
...
             </div>
-            <div data-tooltip="(Die Size: {{secondary.dice}})">
-				{{pfuIconAttribute secondary.attribute}}
+            <div data-tooltip="{{localize 'FU.DieSize'}}: {{secondary.dice}}">
+				{{pfuIconAttribute secondary.attribute flexDir='row' gap='4px' borderless=true size='xs'}}
                 <label class="detail">{{secondary.result}}</label>
             </div>
-            <div {{#if modifiers}}data-tooltip="{{#each modifiers}}{{localize label}}: {{value}}{{#unless @last}}<br>{{/unless}}{{/each}}" {{/if}}>
-				{{pfuBadge 'mod' label='FU.ModAbbr' }}
+            <div data-tooltip="{{#if modifiers}}{{#each modifiers}}{{localize label}}: {{value}}{{#unless @last}}<br>{{/unless}}{{/each}}{{else}}{{localize 'FU.AccuracyCheckBaseAccuracy'}}: 0{{/if}}"">
```
Shourn: There's a surplus quotation mark at the end of this line.
```suggestion
            <div data-tooltip="{{#if modifiers}}{{#each modifiers}}{{localize label}}: {{value}}{{#unless @last}}<br>{{/unless}}{{/each}}{{else}}{{localize 'FU.AccuracyCheckBaseAccuracy'}}: 0{{/if}}">
```

templates/common/icons/badge.hbs
```diff
@@ -1,13 +1,14 @@
 <div class='fu-icon__container flex-group-center'
-	 style="flex-direction: column;"
-	 data-tooltip="{{localize label}}">
+	 style="flex-direction: {{#if (ne flexDir undefined)}}{{flexDir}}{{else}}column{{/if}}; {{#if (ne gap undefined)}}gap: {{gap}};{{/if}}">
```
Shourn: I'm pretty sure this can be simplified.
`{{#if flexDir}}` should (mostly) have the same semantics as the additional `ne` check.
If you want to include `0` as a valid value you can do `{{#if gap includeZero=true}}`

templates/common/icons/badge.hbs
```diff
@@ -1,13 +1,14 @@
...
 	<div class='fu-icon__badge'>
 		<i class='fu-icon--{{size}} {{icon}}'></i>
 		{{#if (ne value undefined)}}
 			<span class="fu-icon__badge-overlay">{{value}}</span>
 		{{/if}}
 	</div>
 	{{#unless compact}}
-		<span class="fu-icon__badge-label">{{localize label}}</span>
+		<span class="fu-icon__badge-label" {{#if (ne fontSize undefined)}}style="font-size: var(--font-size-{{fontSize}});"{{/if}}>
```
Shourn: Same as above.

Shourn (changes requested):

Shourn (approved):

PR #1247, opened 2026-06-21
Bugfix: Weapon table allows creating normal weapons when using technospheres

Shourn:
- only allow creating custom weapons when the technospheres optional rule is active


Frozedon (changes requested):
Confused by this change. The Alternative Technosphere Rules of "Mnemospheres Only" allows for the use of non-custom weapons. So, I don't see the need to automatically restrict to only Custom Weapons if Technospheres are enabled.

Shourn:
You are absolutely right, I completely forgot that optional rule existed.
Don't mind me. 😅

PR #1254, opened 2026-06-26
Initial Foundry V14 compatibility

Shourn:
Initial v14 compatibility.
Even though both ActiveEffects and PseudoDocuments appear to be working for the easy cases I still think there are new edge cases that need to be found and handled.


Unarekin (approved):
Loading a copy of my game's world in v14 did reveal some additional errors, but I'll be setting up separate issues for those shortly

PR #1264, opened 2026-06-27
Register Extended Classes to the SkillLikeItemHelper

### LINKED ISSUE #1256: Skills Don't Show Description When Performed
**Bug Description:**
Skills are not showing any description text or any other effects when performed. Occurs with MiscAbility as well.

**Steps to Reproduce:**
1. Create a PC.
2. Give that PC any non-weapon attack Skill, like Flash of Insight.
3. Perform the Skill.
4. Observe the Chat Card is missing everything except the name.

**Expected behavior:**
Skills should show properly on the Chat card.

Image

Frozedon:
The SkillDataModel, HeroicSkillDataModel, and MiscAbilityDataModel were not being registered to the SkillLikeItemHelper. So, they were being ignored when it went to add renderData of their displays.


Shourn:
I'd prefer if we treated the cause rather than the symptom. Please have a look at my suggested changes.
It already checked for subclasses for `accuracy` checks but I appear to have tunnel visioned and forgotten to also adjust the `attribute` and `display` rendering functions accordingly.

Shourn (approved):

PR #1267, opened 2026-06-27
Fixed Effect Button Remove And Hover Text

### LINKED ISSUE #1261: EFFECT Inline has Incorrect Hover Text Instructions
**Bug Description:**
When you hover over an EFFECT inline button, the hover text has a description that is incorrect. It claims "CTRL+Click disables the effect" which doesn't work and there is no instruction for "Right Click" to view the effect directly.

**Steps to Reproduce:**
1. Create a PC.
2. Give that PC the Hawkeye Skill.
3. Perform the Hawkeye Skill.
4. Hover your mouse over the "Hawkeye" Effect inline button in the Skill's description.
5. Observe the hover text.

**Expected behavior:**
The hover text should be accurate to the inline button's functionality. 

Image

Frozedon:
- Added instruction for examining effect with Right Click to hover text.
- Fixed effect button CTRL+Click function of removing its effect from the actor.
- Plugged some holes in the status to effects that were incorrectly making an Array of statuses instead of a Set.

### DIFF
--- a/module/pipelines/effects.mjs
+++ b/module/pipelines/effects.mjs
@@ -381,7 +381,7 @@ export async function createStatusEffect(actor, statusEffectId, sourceInfo, conf
 		const instance = await ActiveEffect.create(
 			{
 				...statusEffect,
-				statuses: [statusEffectId],
+				statuses: new Set([statusEffectId]),
 				flags: createEffectFlags(statusEffect, sourceInfo, statusEffectId),
 			},
 			{ parent: actor },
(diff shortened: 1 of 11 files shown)


module/pipelines/effects.mjs
```diff
@@ -458,13 +458,13 @@ function removeEffect(document, source, effect) {
 	const existingEffect = document.effects.find(
 		(e) =>
 			e.getFlag(SYSTEM, Flags.ActiveEffect.Temporary) &&
-			e.sourceItem === source &&
+			e.getFlag(SYSTEM, Flags.ActiveEffect.Source)?.name === source.name &&
```
Shourn: The matching by name does not appear to work in all cases, I suggest using fuid instead.
````suggestion
e.getFlag(SYSTEM, Flags.ActiveEffect.Source)?.fuid === source.fuid &&
````

Shourn (changes requested):

Shourn (approved):

PR #1269, opened 2026-06-27
Updated Magiseeds and Haunting Tone Effects

Frozedon:
- Updated all damage change Magiseed effects to add "Ignore Resistances" trait to damage.
- Added Haunting Tone effects to ignore their resistances when they take damage of the given type.


src/packs/skills/effect_Effect__Haunting_Tone__Poison__s4vj7Z41rWYs1JH2.json
Shourn, line 1: There is no "Key" associated with poison damage in the rules.
Should we still include this?

Shourn (changes requested):

src/packs/skills/effect_Effect__Haunting_Tone__Poison__s4vj7Z41rWYs1JH2.json
Frozedon: That's true. I'll remove that

Shourn (approved):

Issue #1294, opened 2026-07-07
Execute Macro Action failing to activate with End of Combat trigger


yoshisman8:
**Bug Description:**
It appears that when one uses the Combat trigger, sets it to End of Combat, and then performs the Execute Macro action, the system will throw out an error and refuse to execute the macro because it is looking for `context.source`, but `context.source` is null, and instead of just executing them with `actor` and `token` being null, it just refuses to execute at all!

<img width="1648" height="747" alt="Image" src="https://github.com/user-attachments/assets/96436e72-a3dc-48a2-8fe1-2012011f4ed9" />

**Steps to Reproduce:**
1. Create a feature with a rule element similar to this
2. <img width="512" height="564" alt="Image" src="https://github.com/user-attachments/assets/b4475e43-46d7-4d45-800c-47ccdf89271a" />
3. have that character enter and start combat, then end the combat

**Expected behavior:**
Even if the actor or token fields are null, it should still execute the macro with what context it *does* have, rather than just refuse to run the macro at all!

yoshisman8:
<img width="457" height="227" alt="Image" src="https://github.com/user-attachments/assets/0ff37a66-1b1b-4923-beb4-0fb98d5d5fbb" />
Literally doing this would fix it quickly... actually hold that thought

PR #1295, opened 2026-07-07
Fixed Issue #1294

yoshisman8:
Added basic null checks to prevent macros from never executing when context.source is undefined/null


Frozedon:
New PR #1303 of this targeting the proper hotfix branch

PR #1297, opened 2026-07-07
Sidebar QoL updates

Unarekin:
- Adds some useful links to the header of the sidebar application
- Updates the header text to use the system's configured `title` property
- Adds the ability to have tooltips on sidebar buttons

image

<br><br>

> [!IMPORTANT]  
> This PR uses the `_loc` helper function added in v14, which will not work in v13.


templates/ui/sidebar-menu.hbs
```diff
@@ -1,19 +1,39 @@
...
-  {{/each}}
\ No newline at end of file
+{{#each tools as |group|}}
+{{#if group.visible}}
+<section class="pfuTools-sidebar__{{group.id}} pfuTools-sidebar__section flexcol form-group {{group.expandedClasses}}">
+  <h4 class="divider pfuTools-sidebar__header">{{localize group.label}}</h4>
+  {{#each group.tools as |tool|}}
+  {{#if tool.visible}}
+  <button type="button" data-action="clickToolButton" data-group="{{group.id}}" data-tool="{{@key}}"
+    class="{{tool.expandedClasses}}" {{#if tool.tooltip}} data-tooltip-html="{{tool.tooltip}}" {{/if}}>
```
Shourn: Wouldn't it be more consistent to localize the tooltip here in the template, same as `group.label` and `tool.label`?

Shourn (changes requested):
Also waiting on discussion results related to `_loc`.

templates/ui/sidebar-menu.hbs
Unarekin: I'd used `data-tooltip-html` with the intent to allow for providing tooltips with HTML content for, like, previews or the like similar to how scene controls have theirs.
I'm wary of potentially passing a block of HTML through the `localize` helper, or of having to stick a bunch of HTML into the localization files themselves

Shourn: Aren't we doing that already? Pretty sure there's a bunch of `<strong></strong>` in the localization texts.

Unarekin: Typography stuff is one thing, but if someone wants to go and stick images or whatnot, it feels wrong having all that in localization strings.

But in either case, we can pass the tooltip through `localize` in the template without breaking things probably, and I'll make that change here in a sec

Shourn (approved):

PR #1308, opened 2026-07-08
Check Rule Predicate Succeeds When Result is Equal

### LINKED ISSUE #1305: Flash of Insight Triggers Only On 14+
**Bug Description:**
The Flash of Insight Rule Trigger only triggers on results of 14 or greater when it should actually trigger at 13.

**Steps to Reproduce:**
1. Create a PC.
2. Give that PC the Flash of Insight Skill.
3. Perform Study actions until you roll a result of 13.
4. Observe that the Skill's rule element didn't trigger.

**Expected behavior:**
Flash of Insight should trigger on any Study Check result of 13 or higher.

Image

Image

Frozedon:
The Check Rule Predicate should succeed when the outcome number given equals the result.

### DIFF
--- a/module/documents/effects/predicates/check-rule-predicate.mjs
+++ b/module/documents/effects/predicates/check-rule-predicate.mjs
@@ -184,7 +184,7 @@ export class CheckRulePredicate extends RulePredicateDataModel {
 		}
 
 		// Check result
-		if (this.result != null && check.result <= this.result) {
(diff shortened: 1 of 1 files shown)


Shourn (approved):
The UX of this field leaves a lot to be desired 😕 From just looking at the UI it may well mean "exactly this result".

PR #1314, opened 2026-07-13
fix:Some CSS updates to ActiveEffect sheet

### LINKED ISSUE #1313: ActiveEffect sheet tabs partially covered
**Bug Description:**
The tabs for the ActiveEffect sheet get scootched up and are partially covered by the header above them.

Image

Unarekin:
- Adds CSS to remove `margin-top` from ActiveEffect sheet top tabs
- Removes `box-shadow`, `border`, and `border-radius` from the document selection button that's added to formGroups generated from `DocumentUUIDField`s
- Replaces the automatically generated formGroup for effects' `origin` property with a text input, as it is always disabled and having the selection button is unnecessary and potentially confusing

So far as I can tell, we aren't generating formGroups for any other `DocumentUUIDField`, but the CSS is included here in case we do in the future or I missed one somewhere, to hopefully have them appear in a less-janky way.

Before:
image

After:
image

### DIFF
--- a/module/documents/effects/active-effect-config.mjs
+++ b/module/documents/effects/active-effect-config.mjs
@@ -194,10 +194,7 @@ export class FUActiveEffectConfig extends foundry.applications.sheets.ActiveEffe
 
 		context.expandedRuleElements = this.#expandedRuleElements;
 
-		if (context.fields?.origin instanceof foundry.data.fields.DocumentUUIDField && context.source?.origin) {
-			const originDoc = await fromUuid(context.source.origin);
-			context.originName = originDoc?.name ?? '';
-		}
+		context.originName = this.document.sourceName;
 		return context;
 	}
 
--- a/module/documents/effects/active-effect-config.mjs
+++ b/module/documents/effects/active-effect-config.mjs
@@ -193,6 +193,11 @@ export class FUActiveEffectConfig extends foundry.applications.sheets.ActiveEffe
 		}
 
 		context.expandedRuleElements = this.#expandedRuleElements;
+
+		if (context.fields?.origin instanceof foundry.data.fields.DocumentUUIDField && context.source?.origin) {
+			const originDoc = await fromUuid(context.source.origin);
+			context.originName = originDoc?.name ?? '';
+		}
 		return context;
 	}
 
--- a/styles/css/global/sheet.css
+++ b/styles/css/global/sheet.css
@@ -1138,6 +1138,7 @@
 
 .projectfu.active-effect-sheet .top-tabs {
 	margin-top: 0;
+	margin-inline: 0;
 }
 
 .projectfu prose-mirror {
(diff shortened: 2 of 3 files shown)


module/documents/effects/active-effect-config.mjs
```diff
@@ -193,6 +193,11 @@ export class FUActiveEffectConfig extends foundry.applications.sheets.ActiveEffe
 		}
 
 		context.expandedRuleElements = this.#expandedRuleElements;
+
+		if (context.fields?.origin instanceof foundry.data.fields.DocumentUUIDField && context.source?.origin) {
```
Shourn: Unless you want to _specifically_ support `origin`s in compendium packs this whole block could be simplified to `context.originName = this.document.sourceName`

styles/css/global/sheet.css
```diff
@@ -1130,6 +1130,16 @@
...
 }
 
+.projectfu document-tags button {
+	box-shadow: none !important;
+	border: none !important;
+	border-radius: 0 !important;
+}
+
+.projectfu.active-effect-sheet .top-tabs {
+	margin-top: 0;
```
Shourn, line 1140: The way the navigation extends right up to the edge of the app looks weird to me.
I propose an additional `margin-block: 0` to align it with the rest of the app content.

Shourn:
In `styles/css/global/theme.css` on line 430 we hard unset the flex gap specifically for the active effect sheet.
I'd like your input on whether we could just remove that.

Shourn (changes requested):

module/documents/effects/active-effect-config.mjs
Unarekin: It's actually partly a vestigial thing, I was originally swapping `context.fields.origin` with a `StringField` to let the `formGroup` helper handle building the HTML, but that felt *too* dirty and I wasn't sure what sort of side effects it might have

That said, supporting origins in compendium packs sounds like a desirable thing?  Or am I mistaken there

styles/css/global/sheet.css
Unarekin: Oh yeah, I'm all for that.
But turns out it's actually `margin-inline: 0`

module/documents/effects/active-effect-config.mjs
Shourn: I'm mostly questioning the actual utility of referencing an origin in a compendium.
As I understand the intent of `origin` the field is meant to link effects to (mostly) their actor of origin to allow "end of next turn" semantics.
Since compendium entries can never participate in actual play I don't think there is much utility here.
At worst it would put more strain on the clients by loading more compendium data.

Unarekin: Fair enough

Shourn (approved):

PR #1332, opened 2026-07-29
Bugfix: Registry field initialization

Shourn:
- add logic to deal with how foundry calls the same method with different options at different points in time

@Unarekin please check if this fixes the issues you encountered


Unarekin:
Still seeing the same issue
image

Unarekin (approved):

PR #1335, opened 2026-07-31
Per-Arcanum Pulse Option

### LINKED ISSUE #1329: Per-Arcanum Pulse Toggle
**User Story:**
As a player/GM I sometimes may want to use core and variant Arcanist in the same game, and having the Pulse option for core Arcana can clutter the Arcanum feature sheet with unnecessary information.

It may be worth introducing a toggle on the Arcanum data model itself, so it can be enabled/disabled on a per-Arcanum basis.

Unarekin:
- Adds the ability to enable/disable Pulse feature per-Arcanum
- Some layout changes to the Arcanum sheet so it isn't so squished
- Migration step to try and intelligently determine if a given Arcanum should have Pulses enabled
- Removes `optionArcanumPulse` setting

<table>
  <tr>
    <th>Without Pulse</th>
    <th>With Pulse</th>
  </tr>
  <tr>
    <td>
      image
    </td>
    <td>
      image
    </td>
  </tr>
</table>


Shourn:
Please have another look at #1332 and see if the newest changes fixed the migration issue.
I'd rather have the migration required for this feature be specific to `ArcanumDataModel` than on the item level.

Unarekin:
> Please have another look at #1332 and see if the newest changes fixed the migration issue. I'd rather have the migration required for this feature be specific to `ArcanumDataModel` than on the item level.

I'd also rather have it on the data model.  A quick check shows that `migrateData` *is* being called on the data model, though still fails when pulse is undefined.

*But* it means fixing that is just adding a migration to the data model -- though we'll need to account for partial updates that don't include those fields.

I know the current code in this PR is good for the migration, so I'll go ahead and move it over to the data model with the expectation that it getting called is fixed in #1332

module/documents/items/classFeature/arcanist/arcanum-migrations.mjs
```diff
@@ -0,0 +1,17 @@
+import { SYSTEM } from '../../../../helpers/config.mjs';
+import { Flags } from '../../../../helpers/flags.mjs';
+
+function migrateEnablePulse(source) {
+	if (foundry.utils.getProperty(source, Flags.migratedEnablePulse)) return;
+
+	if (foundry.utils.hasProperty(source, 'system.data.pulse') && !foundry.utils.hasProperty(source, 'system.data.enablePulse')) {
+		foundry.utils.setProperty(source, 'system.data.enablePulse', !!source.system.data.pulse);
+		foundry.utils.setProperty(source, `flags.${SYSTEM}.${Flags.migratedEnablePulse}`, true);
+	}
```
Shourn: This won't work.
The migration is called with exactly the source data of the `ArcanumDataModel`, so there is no way to get access to the flags.

I suggest the following solution:
```suggestion
if (!('enablePulse' in source) && source.pulse) {
    source.enablePulse = true;
}
```

module/helpers/flags.mjs
```diff
@@ -65,6 +65,7 @@ export const Flags = Object.freeze({
 		ScaleIncomingDamage: 'scaleIncomingDamage',
 	}),
 	ClassConverted: 'classConverted',
+	migratedEnablePulse: 'migratedEnablePulse',
```
Shourn: Likely not needed, since the migrations can't actually use it.

Shourn (changes requested):

module/documents/items/classFeature/arcanist/arcanum-migrations.mjs
Unarekin: Bleh, was hoping to prevent running the migration multiple times just in case it causes issues down the line

What issues?  Dunno!
(the only likely one I can think of would be if enable pulse *should* be enabled, but there happens to be no actual content in the pulse field, which this setup would not cause the enablePulse option to be changed from true to false anyway)

module/helpers/flags.mjs
Unarekin: Definitely not needed, the migrator is the only place that uses it.

module/documents/items/classFeature/arcanist/arcanum-migrations.mjs
Shourn: The proposed migration should only run for arcanums created _before_ these changes were implemented and should stop running as soon as the user changes anything on the sheet.
Until the user changes something in the sheet the source data will not contain the `enablePulse` field and the migration will run, but if they edit the sheet all the inputs get committed into the database including the `enablePulse` checkbox and the migration will stop running.
In the meantime the same logic as your migrations is applied: if there is a pulse description present `enablePulse` is `true`, otherwise it will get initialized to `false` by the datamodel.

Shourn (approved):

PR #1336, opened 2026-07-31
feat:Adding variant breach, pulverizing strike, greater akromorphosis

### LINKED ISSUE #1319: Variant Breach, Pulverizing Strike, Greater Akromorphosis
With the Bestiary being finalized and all official-like, this means the variant class skills included on p 345 fall under the Third Party License.

These skills are meant for games that aren't using equipment for NPCs

Image

Image

Unarekin:
Adds variant Breach, Greater Akromorphosis, and Pulverizing Strike from the Bestiary


Frozedon (changes requested):
Some minor changes
- **Breach**. Accuracy Defense needs to be set to "-", that way it doesn't override the weapon's defense targeting.
- **Breach**. Passive Effect's Rule Element "Calculate Damage" needs to set "Local" option to checked/true. Or it will occur on any damage calculation, not just from Breach.
- **Breach**. Would suggest to use the "Change Traits" action instead, since you're just adding the "IgnoreVulnerabilities" trait.
- **Pulverizing Strike**. The passive effect doesn't seem to do anything, probs get rid of it since we don't have the capability to up a SL with an effect anyway

Unarekin:
> Some minor changes
> 
> * **Breach**. Accuracy Defense needs to be set to "-", that way it doesn't override the weapon's defense targeting.
> * **Breach**. Passive Effect's Rule Element "Calculate Damage" needs to set "Local" option to checked/true. Or it will occur on any damage calculation, not just from Breach.
> * **Breach**. Would suggest to use the "Change Traits" action instead, since you're just adding the "IgnoreVulnerabilities" trait.
> * **Pulverizing Strike**. The passive effect doesn't seem to do anything, probs get rid of it since we don't have the capability to up a SL with an effect anyway

~~It's using Modify Damage specifically because the variant Breach ignores vulnerabilities for *non-Physical* damage types, and Change Traits doesn't look to allow for filtering based on damage type~~
I'm a dummy and wrong

Frozedon (changes requested):
Seems Pulverizing Strike and Breach got duplicate jsons with different ids. Need to delete the duplicates

Frozedon (approved):

PR #1359, opened 2026-08-29
Register standard FU fonts

Unarekin:
Adds the following fonts to the list of available fonts within Foundry:
- Credit Valley
- PFU Sans
- Antonio
- FU Icons

Also adds a snippet to _remove_ the styling from the prosemirror drop-down for the FU Icon font specifically, so it is actually legible.

image


Shourn (approved):
I like the CSS parsing trick, very nice!

PR #1367, opened 2026-09-05
fix: Remove Foundry v14 deprecation warnings for ActiveEffect modes

teo-cay:
I attempted to remove the v14 deprecation warnings for active effect modes

The old CONST.ACTIVE_EFFECT_MODES constant is deprecated (removal in v16) and triggers a console warning on every access. These were replaced with the corresponding string type. 

I'm not sure if PseudoActiveEffect need a migration after this or not. I can look into that if its needed.

Additionally I discoverd an error in Foundry v14.363 that caused the following error. 
```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'handler')
    at ChatMessage.applyMode (foundry.mjs:48538:21)
    at ChatMessage.applyMode (foundry.mjs:48568:22)
    at ChatMessage._preCreate (foundry.mjs:48919:37)
    at async #preCreateDocumentArray (foundry.mjs:80409:29)
    at async ClientDatabaseBackend._createDocuments (foundry.mjs:80361:5)
    at async ChatMessage.create (foundry.mjs:15107:21)
    at async FUChatBuilder.create (chat-builder.mjs:232:3)
    at async performCheck (checks.mjs:589:2)
    at async ActionHandler.promptHinderCheck (action-handler.mjs:126:3)
    at async FUStandardActorSheet.PerformAction (actor-standard-sheet.mjs:912:3)
```
Because messageMode expects a string and not a boolean. Note that this error doesn't happen on 14.367+ but that is because foundry added a check that does the parsing with a boolean is sent, but from what I've gathered the correct way is to send the message mode from settings. It is optional but will give compatibility for pre 14.367 builds

### DIFF
--- a/module/helpers/chat-builder.mjs
+++ b/module/helpers/chat-builder.mjs
@@ -225,7 +225,7 @@ export class FUChatBuilder {
 		// OPTION: Roll data
 		if (this.#rolls) {
 			chatMessage.rolls = this.#rolls;
-			options.messageMode = true;
+			options.messageMode = game.settings.get('core', 'messageMode');
 		}
 
 		// Render to chat
--- a/module/documents/effects/active-effect.mjs
+++ b/module/documents/effects/active-effect.mjs
@@ -97,7 +97,7 @@ Hooks.on('preCreateActiveEffect', (effect, options, userId) => {
 
 Hooks.on('updateActiveEffect', (effect, changes, options, userId) => {
 	if (game.userId === userId && effect.target instanceof FUActor && effect.target.canUserModify(game.user, 'update')) {
-		if (effect.changes.some((change) => change.key.startsWith('system.resources.hp'))) {
+		if (effect.system.changes.some((change) => change.key.startsWith('system.resources.hp'))) {
 			effect.target.applyCrisis();
 		}
 		if ('disabled' in changes) {
--- a/module/documents/effects/pseudo-active-effect.mjs
+++ b/module/documents/effects/pseudo-active-effect.mjs
@@ -16,7 +16,7 @@ class BasePseudoActiveEffect extends PseudoDocument {
 				new fields.SchemaField({
 					key: new fields.StringField({ required: true }),
 					value: new fields.StringField({ required: true }),
-					mode: new fields.NumberField({ required: true, nullable: false, integer: true, initial: CONST.ACTIVE_EFFECT_MODES.ADD }),
+					type: new fields.StringField({ required: true, initial: 'add' }),
 					priority: new fields.NumberField(),
 				}),
 			),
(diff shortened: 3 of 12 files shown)


Unarekin:
I'm not a big fan of moving over towards using a set of magic strings for this, though I don't see anywhere where these values are already defined in a handy enum, so this may very well be a place to define that for future-proofing.

teo-cay:
I agree on the magic strings but it seems that there is no current constant for that currently in core. Having an internal mapping for this system is a possibility but it seemed a bit overkill for this specifically, but I could do that if that’s what you meant

Shourn:
Once again, thanks for the contribution.
I did some additional refactoring:

First and most importantly, I realigned the data model of `BasePseudoActiveEffect` with `BaseActiveEffect`.
I must've missed this, back when I aligned `Actor` and `Item`.
Then I changed `type: 'custom'` to `type: FU.changeTypes.apply`.
The `custom` change type is effectively dead and buried, the foundry devs want everyone to use bespoke change types in the future.
I also changed places where `effect.changes` was used into `effect.system.changes`.
For now the old version would have worked fine, but since `effect.system.changes` is now the proper location for this data I decided that consistently referring to `effect.system.changes` would reduce confusion and the chance for errors.

> I'm not a big fan of moving over towards using a set of magic strings for this, though I don't see anywhere where these values are already defined in a handy enum, so this may very well be a place to define that for future-proofing.

The core change types are available as the keys of `CONST.ACTIVE_EFFECT_CHANGE_TYPES` and are mapped to their default priorities.
While I don't like that the Foundry devs are using magic strings across their codebase, unilaterally defining constants on our end doesn't really sit right with me either.
I'd say we just have to live with some magic strings.

module/helpers/chat-builder.mjs
```diff
@@ -225,7 +225,7 @@ export class FUChatBuilder {
 		// OPTION: Roll data
 		if (this.#rolls) {
 			chatMessage.rolls = this.#rolls;
-			options.messageMode = true;
+			options.messageMode = game.settings.get('core', 'messageMode');
```
Shourn, line 228: I do understand why this is the more robust version in respect to older Foundry 14 versions.
But on the other hand I chose `options.messageMode = true` for a reason.
The reason was that I wanted Foundry to determine the active `messageMode` on our behalf, without us having to query the setting ourselves.
As far as I can tell Foundry has no central documented list for available `core` settings and as such I'd rather treat them as implementation details than public API.

Don't take this as an instruction to change it back, I just wanted to explain why it was the way it was before.

teo-cay: I agree with your reasons. That’s why I flagged it as an optional thing. Some systems pin the minimum core version to a specific minor release so that could also be an option to avoid this.

I just discovered this error randomly because I had the “wrong” version installed and the error is not very user friendly.

teo-cay:
Thanks for the explanation, that’s why I didn’t touch those effects as I couldn’t really understand why the pseudo ones where different than the rest

PR #1368, opened 2026-09-07
Chore: cleanup dead gulpfile and scss references

teo-cay:
The `CONTRIBUTING.md` file contains outdated commands and references to compiling css that is outdated. I got confirmation through discord that these are dead files that can be removed.

This will help new contributors with setting up their dev environments easier.

I tried to bring this in line with the wiki's `Development-Environment.md` but I can't figure out how to raise a PR for the wiki, I can share the file here if you want, but it should be enough to remove references to `npm run build` and `npm run watch`


Unarekin:
Just a quick note to mention that you couldn't figure out how to do a PR against the wiki because you just straight up can't.  The wiki repo isn't handled in the same way the "real" one is.

teo-cay:
Ok I’ll create an issue for the wiki update then :)

Shourn (approved):

Issue #1127, opened 2026-05-17
DataModel update special operation key deprecation


Unarekin:
Foundry v14 deprecates the `==` and `-=` operators for DataModel updates to replace and remove keys with [_replace](https://foundryvtt.com/api/classes/foundry.data.operators.ForcedReplacement.html) and [_del](https://foundryvtt.com/api/classes/foundry.data.operators.ForcedDeletion.html), respectively.

Support for these keys will be removed in v16

https://github.com/foundryvtt/foundryvtt/issues/13090

PR #1371, opened 2026-09-09
Refactor/data model operations

teo-cay:
Remove  deprecated operators for DataModel updates as described in #1127 

The only old key I found was in `module/ui/npc-profile.mjs`

For consistency's sake I also updated the use of `ForcedReplacement` as per the api docs linked in the issue. 

`module/systems/pressure-system.mjs` whas throwing a TypeError, that should be fixed with the correct use of the constructor.

### DIFF
--- a/module/ui/npc-profile.mjs
+++ b/module/ui/npc-profile.mjs
@@ -357,8 +357,7 @@ export class NpcProfileWindow extends FUApplication {
 			if (changedAdversary.revealed.traits) {
 				changedAdversary.revealed.traits = changedAdversary.revealed.traits.filter((t) => traitsArray.includes(t));
 				if (changedAdversary.revealed.traits.length === 0) {
-					delete changedAdversary.revealed.traits;
-					changedAdversary.revealed['-=traits'] = null;
+					changedAdversary.revealed.traits = new foundry.data.operators.ForcedDeletion();
 				}
 			}
 			await party.parent.update({ system: { adversaries: adversaries } });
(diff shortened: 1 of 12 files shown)


module/ui/npc-profile.mjs
```diff
@@ -357,8 +357,7 @@ export class NpcProfileWindow extends FUApplication {
 			if (changedAdversary.revealed.traits) {
 				changedAdversary.revealed.traits = changedAdversary.revealed.traits.filter((t) => traitsArray.includes(t));
 				if (changedAdversary.revealed.traits.length === 0) {
-					delete changedAdversary.revealed.traits;
-					changedAdversary.revealed['-=traits'] = null;
+					changedAdversary.revealed.traits = new foundry.data.operators.ForcedDeletion();
```
Shourn: To not cause any undue confusion we should probably use the factory method everywhere.

```suggestion
					changedAdversary.revealed.traits = foundry.data.operators.ForcedDeletion.create();
```

Shourn (changes requested):
Could I impose on you to replace uses of `new ForcedDeletion()` with `ForcedDeletion.create()` as well?
Being consistent in our use of the factory method across all operators will reduce the confusion and potential for wrong usage in the future.

That being said, thank you for reading the docs and noticing something I completely missed.

module/ui/npc-profile.mjs
teo-cay: Sure I can, the reason I didn't replace all is that `new ForcedDeletion()` is not deprecated. It only inherits the generic DataFieldOperator.create(value) so as far as I can see `ForcedDeletion.create()` is functionally identical to` new ForcedDeletion()` 
`ForcedReplacement` overrides `.create()` which has somme added functionality so it is not the same case.

But I can see the point of using factory method everywhere to make it more consistent.

Many of the `...table-renderer.mjs` files use also `new ForcedDeletion()`, should I change those too?

Shourn: Yes, I would replace them everywhere.
The table renderers uses `foundry.utils.mergeObject` with the `applyOperators` option, so operators should work just like when used with `Document#update`.

teo-cay: done! I think I got all of them now

Shourn (approved):
Looks good to me.
Thank you very much.

Issue #1202, opened 2026-06-06
Roll NPCs button


hiherp:
As a GM, I would like to make use of the at least one of the two buttons on the Combat Tracker, Roll NPCs (and Roll All), or something functionally equivalent.

When clicking Roll NPCs, a list of opposing actors would appear, allowing me to select a leader. Once confirmed, the initiative checks for all relevant creatures would be resolved automatically.

I do not see myself using Roll All very often, even in other systems, but it may still be useful for some groups or situations. Whether it remains as-is, is removed, or instead prompts the GM to roll separately for both sides is not particularly important to me. Personally, I think it best to either be removed or to function in a similar way.

<img width="257" height="102" alt="Image" src="https://github.com/user-attachments/assets/ab365a47-f162-47ef-a81e-b3077393802f" />

Unarekin:
If nothing, Roll NPCs probably ought to be removed as RAW, NPCs don't roll Initiative at all.

PR #1381, opened 2026-09-14
Fix/remove npc rollinit

teo-cay:
Fix for #1202 

Split the changes in two commits:
 - The first one simply disables the roll npcs button as suggested in the issue.
 - The second one (optional) adds functionality to the roll all button, prompting the gm to select a leader and then triggering the regular group-check initiative flow. I figured we could add something to that one instead of just removing it.


Shourn (changes requested):
While I like this change in theory, I have a bit of an issue with one specific part of how it works in practice here.

I am a big believer in the players always rolling dice for their characters.
With this implementation the GM gets to roll the final initiative check, with the player of the leading character being effectively "locked out" from participating.
To remedy this you'll need to send a socket message to the leading character's player with instructions to start an initiative check.
Please have a look at `socket.mjs` to see how we handle socket messages.

teo-cay:
Right, I didn’t think about that (I’m a forever GM hehe)
I’ll look into the socket messages

teo-cay:
Now the selected player should be able to continue with the regular initiative flow as if they had started it.

I added a check so that if there are no controlling players the GM will then carry on with the roll. 

I'm not sure what is supposed to happen if there are multplie owners of an actor. It doesn't make sense to me to send the prompt to start the initiative check to everyone, should it then fail? or the gm complete it?

I'd appreciate any input on how that case should be handled :)

module/socket.mjs
```diff
@@ -35,6 +37,7 @@ export const MESSAGES = Object.freeze({
 	RequestEndTurn: 'requestEndTurn',
 	RequestTrade: 'requestTrade',
 	RequestZenitTransfer: 'requestZenitTransfer',
+	RequestRollAll: 'requestRollAll',
```
Unarekin: Would be better if this were named in such a way that it's more clear that this is for rolling initiative for everyone

module/socket.mjs
```diff
@@ -350,3 +354,14 @@ async function requestEndTurn(combatId, combatantId) {
 		if (combatant) await combat.endTurn(combatant);
 	}
 }
+
+/**
+ * Starts an initiative group check for the actor with the given ID on the receiving client,
+ * so that the leader character's owning player performs the roll themselves.
+ * @param {string} actorId
+ */
+async function requestRollAll(actorId) {
```
Unarekin: I would prefer we avoid adding callback handlers directly in the socket manager where it is feasible (yes I know there are several here already, they're older ones and probably need new homes also) to avoid clutter

This should be able to be shifted over into the combat tracker with the rest of the code for this process.

Unarekin (changes requested):
> Now the selected player should be able to continue with the regular initiative flow as if they had started it.
> 
> I added a check so that if there are no controlling players the GM will then carry on with the roll.
> 
> I'm not sure what is supposed to happen if there are multplie owners of an actor. It doesn't make sense to me to send the prompt to start the initiative check to everyone, should it then fail? or the gm complete it?
> 
> I'd appreciate any input on how that case should be handled :)

In the case of actors with multiple owners, look to see if a user has that actor assigned as their character (it will be in the user's `character` property).  That should be a fairly definitive indicator of ownership

Beyond that, if there are multiple owners, neither has it assigned, and both users are online, I think it is safe to fail out with a warning that it couldn't determine who

module/socket.mjs
teo-cay: ok! I named it atfer the button, something like `requestInitiativeRollAll` is better? I hate naming things

teo-cay:
> > Now the selected player should be able to continue with the regular initiative flow as if they had started it.
> > I added a check so that if there are no controlling players the GM will then carry on with the roll.
> > I'm not sure what is supposed to happen if there are multplie owners of an actor. It doesn't make sense to me to send the prompt to start the initiative check to everyone, should it then fail? or the gm complete it?
> > I'd appreciate any input on how that case should be handled :)
> 
> In the case of actors with multiple owners, look to see if a user has that actor assigned as their character (it will be in the user's `character` property). That should be a fairly definitive indicator of ownership
> 
> Beyond that, if there are multiple owners, neither has it assigned, and both users are online, I think it is safe to fail out with a warning that it couldn't determine who

In the case where no owners are configured I had the roll continue with the GM making the group check, should this be the same for the "couldn't determine who" case? or should that one simply fail?

I guess the case for the GM making the roll is when playing for someone who is offline that session. But wanted to double check

teo-cay:
New implementation looks like this for different scenarios after choosing the leader for the check:

Leader's assigned player online -> That player rolls the lead check
Multiple owners, none assigned -> GM rolls (warning)
Assigned player offline -> GM rolls (warning)
No owners at all -> GM rolls (warning)

The reason for that is that in my experience if the gm selects a leader that fits the above scenarios is usually intentional, and in the rare case that it was a missclick/mistake they can just cancel the roll and go again or have the gm roll instead of erroring out.

Shourn (approved):
LGTM

Unarekin (approved):

